PR #2, opened 2020-10-09
Show errors from contract mutation hooks

danieldunderfelt:
Closes https://github.com/HSLdevcom/bultti/issues/51
Instead of showing the mutation request result error, show errors from the mutation hooks. This is more dynamic and reactive and doesn't require local state for the error.

Also, it should show errors from both the create and update mutations.


Divinare:
LGTM :+1:

PR #5, opened 2020-10-20
Navigation: block redirect with unsaved changes

Divinare:
Relates to: https://github.com/HSLdevcom/bultti/issues/61

Note: Chrome displays the string, "Do you want to leave this site? Changes you made may not be saved." This message can't be changed. Also firefox has its own message.


src/common/input/InputForm.tsx
```diff
@@ -1,7 +1,9 @@
-import React from 'react'
+import _ from 'lodash'
```
danieldunderfelt, line 1: Use `import { what, you, need } from 'lodash'` instead.

src/common/input/InputForm.tsx
```diff
@@ -19,8 +21,17 @@ export type PropTypes = {
 
 const InputForm: React.FC<PropTypes> = observer(
   ({ onDone, onCancel, doneLabel, doneDisabled, fields, children: buttons }) => {
+    let MemorizedPromptUnsavedChanges = useMemo(() => {
```
danieldunderfelt, line 24: Is it really necessary to memoize this component? Also, the uniqueId will change when doneDiabled updates. A better way would probably be to memoize a uniqueId that never changes and mount the component as normal.

src/common/input/InputForm.tsx
```diff
@@ -19,8 +21,17 @@ export type PropTypes = {
...
+      return (
+        <PromptUnsavedChanges
+          uniqueComponentId={_.uniqueId()}
```
danieldunderfelt, line 27: So instead of memoizing this component, you would do:

```
let formId = useMemo(() => uniqueId(), [])
```

And pass formId to the component.

src/common/components/PromptUnsavedChanges.tsx
```diff
@@ -0,0 +1,41 @@
+import _ from 'lodash'
```
danieldunderfelt, line 1: Use import { what, you, need } from 'lodash' instead.

src/common/components/PromptUnsavedChanges.tsx
```diff
@@ -0,0 +1,41 @@
...
+  uniqueComponentId: currentId,
+  shouldShowPrompt,
+}) => {
+  const prevIdRef = useRef(currentId)
+  let [unsavedFormIds, setUnsavedFormIds] = useStateValue('unsavedFormIds')
+  let [, removeUnsavedFormId] = useStateValue('removeUnsavedFormId')
+
+  const prevId: string = prevIdRef.current
+  useEffect(() => {
+    let newUnsavedFormIds = unsavedFormIds ? _.cloneDeep(unsavedFormIds) : []
```
danieldunderfelt, line 21: It would probably be better to just use `[...unsavedFormIds]` here instead of cloneDeep.

src/common/components/PromptUnsavedChanges.tsx
```diff
@@ -0,0 +1,41 @@
...
+    // In case of id has changed, remove old id
+    if (prevId !== currentId) {
+      newUnsavedFormIds = newUnsavedFormIds.filter((id) => id !== prevId)
+      prevIdRef.current = currentId
+    }
+    if (shouldShowPrompt) {
+      newUnsavedFormIds = newUnsavedFormIds ? newUnsavedFormIds.concat(currentId) : [currentId]
```
danieldunderfelt, line 28: Do you mean `push` instead of concat? Also, the condition will never be false as `newUnsavedFormIds` is always an array. Use `newUnsavedFormIds.length !== 0 ? ...` instead. You don't even need a condition, you can just push currentId into newUnsavedFormIds in all cases.

src/common/components/PromptUnsavedChanges.tsx
```diff
@@ -0,0 +1,41 @@
...
+  shouldShowPrompt: boolean
+}
+
+const PromptUnsavedChanges: React.FC<PropTypes> = ({
+  uniqueComponentId: currentId,
+  shouldShowPrompt,
+}) => {
+  const prevIdRef = useRef(currentId)
+  let [unsavedFormIds, setUnsavedFormIds] = useStateValue('unsavedFormIds')
+  let [, removeUnsavedFormId] = useStateValue('removeUnsavedFormId')
```
danieldunderfelt, line 17: Hmm, this is not how I intended `useStateValue` to be used. `setUnsavedFormIds` should serve all needs, also in this case.

src/state/UIStore.ts
```diff
@@ -51,11 +51,21 @@ export const UIStore = (state): UIActions => {
     state.appLoaded = true
   })
 
+  const setUnsavedFormIds = action((unsavedFormIds: string[]) => {
+    state.unsavedFormIds = unsavedFormIds
+  })
+
+  const removeUnsavedFormId = action((idToRemove: string) => {
+    state.unsavedFormIds = state.unsavedFormIds.filter((id) => id !== idToRemove)
+  })
```
danieldunderfelt, line 60: Creating a new action for this seems unnecessary, as you can just as easily use the `setUnsavedFormIds` action.

src/common/components/PromptUnsavedChanges.tsx
```diff
@@ -0,0 +1,41 @@
...
+      prevIdRef.current = currentId
+    }
+    if (shouldShowPrompt) {
+      newUnsavedFormIds = newUnsavedFormIds ? newUnsavedFormIds.concat(currentId) : [currentId]
+    } else {
+      newUnsavedFormIds = newUnsavedFormIds.filter((id) => id !== currentId)
+    }
+    setUnsavedFormIds(newUnsavedFormIds)
+    return () => {
+      removeUnsavedFormId(currentId)
```
danieldunderfelt, line 34: Instead of creating a new action, just write `unsavedFormIds.filter((id) => id !== currentId)`. Also, you should probably not use the cleanup function here and implement the removal of the ID in the effect itself. Then add unsavedFormIds to the dependency array of the effect.

src/common/components/PromptUnsavedChanges.tsx
```diff
@@ -0,0 +1,27 @@
...
+  uniqueComponentId: string
+  shouldShowPrompt: boolean
+}
+
+const PromptUnsavedChanges: React.FC<PropTypes> = ({
+  uniqueComponentId: currentId,
+  shouldShowPrompt,
+}) => {
+  let [unsavedFormIds, setUnsavedFormIds] = useStateValue('unsavedFormIds')
+  let nextUnsavedIds = [...unsavedFormIds]
```
danieldunderfelt, line 15: This should be inside the effect.

Divinare:
Review fixes done ^

src/util/promptUnsavedChanges.ts
```diff
@@ -0,0 +1,40 @@
...
+    if (shouldShowPrompt) {
+      nextUnsavedIds.push(uniqueComponentId)
+    } else {
+      nextUnsavedIds = nextUnsavedIds.filter((id) => id !== uniqueComponentId)
+    }
+    setUnsavedFormIds(nextUnsavedIds)
+  }, [shouldShowPrompt])
+}
+
+export const promptUnsavedChangesOnClick = (clickEvent: React.MouseEvent) => ({
```
danieldunderfelt, line 26: It's not really necessary to use a nested/curried function like this since you always call the second one directly. And if you want to have it like this, you should switch the arguments the other way around so that the FIRST function is called with `unsavedFormIds` and `setUnsavedFormIds` and the second is called with the event.

But I'd rather combine the unsaved form ID arguments into one, called `unsavedFormState`, and add a second argument called something like `onDeclined`, that you call if the `window.confirm` call comes back false. Then you'd call the function simply like this:

```
// No need to restructure the array
let unsavedFormState = useStateValue('unsavedFormIds')
<button onClick={event => promptUnsavedChangesOnClick(unsavedFormState, event.preventDefault)}
```

The function doesn't need to know anything about the event, it just needs a function to call if the user declines the confirm alert. And for the first argument, the convention of state values being `[value, action]` is well established in this project and it doesn't need to be unpacked or restructured for this function.

Divinare: Refactored according to this.

Divinare:
Some forms (e.g. in inspection page) are not showing alerts. Added a refactoring task: https://github.com/HSLdevcom/bultti/issues/64 to fix that issue.

PR #8, opened 2020-11-24
Post inspection WIP

danieldunderfelt:
Everything related to the post-inspection for the UI.


src/common/components/Table.tsx
```diff
@@ -628,7 +639,9 @@ const Table = observer(
       items.length === 0 ||
       (items.length === 1 && Object.values(items[0]).every((val) => !val))
 
-    let width = fluid ? '100%' : Math.ceil(columnWidths.reduce((total, col) => total + col, 0))
+    let width = fluid
```
Divinare, line 642: Would hope that all things related to width would be solved with CSS somehow. I do agree that it might be difficult sometimes with tables

src/report/ListReport.tsx
```diff
@@ -36,25 +39,20 @@ const ListReport = observer(
 
     return (
       <ListReportView>
-        <ReportTableFilters<ItemType>
+        <Table<ItemType>
+          virtualized={true}
+          keyFromItem={(item: any) =>
+            item?.id || item?._id || item?.departureId || item?.registryNr || ''
+          }
+          maxHeight={window.innerHeight * 0.75}
```
Divinare, line 47: Better practice would be having style related code in CSS to have cleaner code.

src/util/round.ts
```diff
@@ -1,13 +1,21 @@
...
+  return parseInt('1'.padEnd(decimals + 1, '0'), 10)
 }
 
-// Round up to three decimals
-export function ceil(number: number): number {
-  return Math.ceil(number * 1000) / 1000
+// Round down
+export function floor(number: number, decimals = 3): number {
+  let decimalsMultiplier = getMultiplier(decimals)
+  return Math.floor(number * decimalsMultiplier) / decimalsMultiplier
```
Divinare, line 8: Use decimals is type :number? That way you don't need getMultiplier() function and you can use decimals value straightly in here

Divinare:
Code review done. Will create new cards of those UI things I mentioned previously.

Testing: opened a pre-inspection page but that calculation button didn't quite work yet. Will try to setup that runner tomorrow that does the job.

PR #9, opened 2020-11-26
83 procurement units page fixes

Divinare:
Closes: https://github.com/HSLdevcom/bultti/issues/83

Note: I moved task "add a remove button to each table row" into our table component fixes card: https://github.com/HSLdevcom/bultti/issues/85


danieldunderfelt (changes requested):
There is already a translation feature in Bultti, found in `src/util/translate.ts`. It uses text defined in the folders `src/text/*`. Please move the help text to the fi.json file in the help folder, define tokens for each hint and use the translate() function to show the hint text.

Divinare:
Review fixes done.

Using translate() now, also put those hint texts under hint. prefix key instead of help.fi.json (removed that file from being used)

danieldunderfelt (approved):
Looks good!

PR #10, opened 2020-11-27
Expandable section can be opened / closed from anywhere of the header

Divinare:
Closes: https://github.com/HSLdevcom/bultti/issues/84

We wanted to implement way to open/close from the whole expandable bar

- we dont want to open/close:
 - from the content of expandable bar
 - from any buttons inside expandable bar, places where this was prevented
    - post-inspections page at UserInspection section)
    - expandable bar's button on the right

### DIFF
--- a/src/common/components/ExpandableSection.tsx
+++ b/src/common/components/ExpandableSection.tsx
@@ -1,6 +1,7 @@
 import React, { CSSProperties, useCallback, useEffect, useState } from 'react'
 import styled, { css } from 'styled-components'
 import { observer } from 'mobx-react-lite'
+import { forOwn } from 'lodash'
 import { ArrowDown } from '../icon/ArrowDown'
 import { SectionHeading } from './Typography'
 
@@ -141,7 +142,22 @@ const ExpandableSection = observer(
   }: PropTypes) => {
     const [expanded, setExpanded] = useState(isExpanded)
 
-    let onChangeExpanded = useCallback(
+    let onClickHeaderRow = useCallback(
+      (e: React.MouseEvent) => {
+        let wasExpandableSectionClicked = false
+        forOwn((e.target as Element).classList, (value) => {
+          if ((value as string).includes('ExpandableSection'))
+            wasExpandableSectionClicked = true
+        })
+
+        if (wasExpandableSectionClicked) {
+          setExpanded((currentVal) => !currentVal)
+        }
+      },
+      [onToggleExpanded]
+    )
+
+    const onClickExpandToggle = useCallback(
       (e: React.MouseEvent) => {
         setExpanded((currentVal) => !currentVal)
         e.stopPropagation()
@@ -160,28 +176,20 @@ const ExpandableSection = observer(
       }
     }, [isExpanded])
 
-    const stopEventPropagation = useCallback((e: React.MouseEvent) => {
-      e.stopPropagation()
-    }, [])
-
     return (
-      <ExpandableBoxView
-        error={error}
-        style={style}
-        className={className}
-        onClick={onChangeExpanded}>
-        <HeaderRow>
+      <ExpandableBoxView error={error} style={style} className={className}>
+        <HeaderRow onClick={onClickHeaderRow}>
           {headerContent && (
             <HeaderContentWrapper expanded={expanded}>
               {typeof headerContent === 'function' ? headerContent(expanded) : headerContent}
             </HeaderContentWrapper>
           )}
-          <ExpandToggle expanded={expanded} onClick={onChangeExpanded}>
+          <ExpandToggle expanded={expanded} onClick={onClickExpandToggle}>
             <ArrowDown width="1rem" height="1rem" fill="var(dark-grey)" />
           </ExpandToggle>
         </HeaderRow>
         {(expanded || !unmountOnClose) && (
-          <ContentWrapper expanded={expanded} onClick={stopEventPropagation}>
+          <ContentWrapper expanded={expanded}>
             {typeof children === 'function' ? children(expanded) : children}
           </ContentWrapper>
         )}
--- a/src/common/components/ExpandableSection.tsx
+++ b/src/common/components/ExpandableSection.tsx
@@ -9,6 +9,7 @@ const ExpandableBoxView = styled.div<{ error: boolean }>`
   margin-top: 1rem;
   border-radius: 0.5rem;
   background: white;
+  cursor: pointer;
 `
 
 export const HeaderRow = styled.div`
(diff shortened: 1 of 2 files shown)


src/inspection/InspectionUsers.tsx
```diff
@@ -62,7 +66,10 @@ const InspectionUsers: React.FC<PropTypes> = observer(({ inspection }) => {
               style={{ marginLeft: 'auto' }}
               buttonStyle={ButtonStyle.SECONDARY}
               size={ButtonSize.SMALL}
-              onClick={refetchRelations}>
+              onClick={(e: React.MouseEvent) => {
+                refetchRelations()
+                stopEventPropagation(e)
+              }}>
```
danieldunderfelt, line 72: I would rather create a new callback, `onRefetchRelations`, which stops the event propagation, than make a separate stopPropagation function.

```suggestion
onClick={onRefetchRelations}>
```

src/inspection/InspectionUsers.tsx
```diff
@@ -52,6 +52,10 @@ const InspectionUsers: React.FC<PropTypes> = observer(({ inspection }) => {
     }
   }, [inspection, user, toggleSubscribed, refetchRelations])
 
+  const stopEventPropagation = useCallback((e: React.MouseEvent) => {
+    e.stopPropagation()
+  }, [])
+
```
danieldunderfelt, line 58: If necessary to stop propagation here, but I would rather have the ExpandableBox component catch and stop the propagation and not have to do that separately in all components which use the expandable box.

```suggestion
  const onRefetchRelations = useCallback((e: React.MouseEvent) => {
    e.stopPropagation()
    refetchRelations()
  }, [refetchRelations])

```

src/common/components/ExpandableSection.tsx
```diff
@@ -155,8 +160,16 @@ const ExpandableSection = observer(
...
+    }, [])
+
     return (
-      <ExpandableBoxView error={error} style={style} className={className}>
+      <ExpandableBoxView
+        error={error}
+        style={style}
+        className={className}
+        onClick={onChangeExpanded}>
         <HeaderRow>
```
danieldunderfelt, line 173: Would it be effective to set the stopPropagation handler on HeaderRow, and have it catch and stop propagation from all buttons in the header? This would make it unnecessary to stop propagation in the button handlers separately.

danieldunderfelt (changes requested):
Check the code comments.

src/common/components/ExpandableSection.tsx
Divinare: This was what I initially wanted but didn't figure out a way. Now I just commited a solution where you don't need to handle stop propagation in button handlers!

src/inspection/InspectionUsers.tsx
Divinare: No longer have to stop propagation here

Divinare:
Review fix done! Now it's only ExpandableSection's responsibility to handle stopPropagations.

Yes, there are two methods (the click handlers onClickHeaderRow, onClickExpandToggle) that call setExpanded(), but I think that way the code is cleaner

src/common/components/ExpandableSection.tsx
```diff
@@ -140,9 +142,28 @@ const ExpandableSection = observer(
   }: PropTypes) => {
     const [expanded, setExpanded] = useState(isExpanded)
 
-    let onChangeExpanded = useCallback(() => {
-      setExpanded((currentVal) => !currentVal)
-    }, [onToggleExpanded])
+    let onClickHeaderRow = useCallback(
+      (e: React.MouseEvent) => {
+        let wasExpandableSectionClicked = false
+        forOwn((e.target as Element).classList, (value) => {
```
danieldunderfelt, line 148: I would rather check that the `tagName` of the target isn't a `button`, `input`, or `a`. The className is not reliable.

src/common/components/ExpandableSection.tsx
```diff
@@ -157,13 +178,13 @@ const ExpandableSection = observer(
...
       <ExpandableBoxView error={error} style={style} className={className}>
-        <HeaderRow>
+        <HeaderRow onClick={onClickHeaderRow}>
           {headerContent && (
             <HeaderContentWrapper expanded={expanded}>
               {typeof headerContent === 'function' ? headerContent(expanded) : headerContent}
             </HeaderContentWrapper>
           )}
-          <ExpandToggle expanded={expanded} onClick={onChangeExpanded}>
+          <ExpandToggle expanded={expanded} onClick={onClickExpandToggle}>
```
danieldunderfelt, line 187: No need to have an event handler here as you are already listening for click events in the parent.

src/common/components/ExpandableSection.tsx
```diff
@@ -157,13 +178,13 @@ const ExpandableSection = observer(
 
     return (
       <ExpandableBoxView error={error} style={style} className={className}>
-        <HeaderRow>
+        <HeaderRow onClick={onClickHeaderRow}>
           {headerContent && (
             <HeaderContentWrapper expanded={expanded}>
```
danieldunderfelt, line 183: Suggestion: move the element type check to a click handler on this element from `onClickHeaderRow` and stopPropagation if necessary?

danieldunderfelt (changes requested):
A few comments and suggestions.

src/common/components/ExpandableSection.tsx
Divinare: If I take this click handler out, then yes it does trigger the parent's click handler. But the problem is className is not 'ExpandableSection' so setExpanded() won't get called.

src/common/components/ExpandableSection.tsx
Divinare: That is not reliable either though, there can still be a child with a className 'ExpandableSection' which type is `div` and then this would get triggered from that. As long as there is no child with className 'ExpandableSection' this isn't a problem. I think the most secure way might be to use 'ref' somehow

Divinare:
Now we count the number of classes found from click event. Couldn't find any other secure way to ensure this

src/common/components/ExpandableSection.tsx
Divinare: Tried having one method with:
```
onClick={(e) => onClickHeaderRow(e)({ isForced: false })}
if (isForced) {
  setExpanded((currentVal) => !currentVal)
  e.stopPropagation()
  return
}
```

But I prefered having two methods instead of passing booleans and having an additional if statement there

src/common/components/ExpandableSection.tsx
danieldunderfelt: TagName is very reliable. We should always make clickable elements with `button`, except in this case where we have to put a click handler on a div. But all clickable elements within the header will always be either button, a, or input. We can add to this list if necessary.
I would say you never have to look at the className in React if you're using styling solutions such as styled-components as we are here.
Check the suggestion in the main thread of this PR.

src/common/components/ExpandableSection.tsx
danieldunderfelt: No, we don't. Check the suggestion in the main thread of this PR.

danieldunderfelt (changes requested):
This is my suggestion for a very simple implementation. I should have done this from the start:

https://github.com/HSLdevcom/bultti-ui/pull/13/files

Two click handlers and one is just for stopping propagation if the clicked element is a button. Checking the className is not necessary.

Divinare:
Lets use the suggestion here: https://github.com/HSLdevcom/bultti-ui/pull/13

With knowing of that conviention of using always <Button /> instead of <div />, that solution is ok

PR #14, opened 2020-12-02
Inspection date range entity & resolvers

Divinare:
**Made a new card: InspectionForms: date options https://github.com/HSLdevcom/bultti/issues/95**
(new card is needed because minStartDate, inspectionStartDate and inspectionEndDate are to be removed)

**Suggestion: new card -> https://github.com/HSLdevcom/bultti/issues/96**

Refactor like this:

```
export function getIsUserAdmin(): boolean {
  let [user] = useStateValue('user')
  return requireUser(user) && user?.role === UserRole.Admin
}
```
(also in backend + other methods too. Move role checking into backend)

**Suggestion 2: new card -> ?**

Validations by using some easilly configurable models like in jore-map-ui:
https://github.com/HSLdevcom/jore-map-ui/blob/develop/src/models/validationModels/lineValidationModel.ts

**Suggestion 3: new card -> ?**

We need a way to pass dates to the calendar that are unselectable. I doubt that the current native date input can pull this off. Have to switch to using some calendary library I think.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -133,7 +133,7 @@ const App: React.FC = observer(() => {
         {requireAdminUser(user) && <InspectionDatePage path="inspection-date" />}
         <UserPage path="user" />
         <OperatorContractsListPage path="contract" />
-        <EditContractPage path="contract/:contractId" />
+        {requireAdminUser(user) && <EditContractPage path="contract/:contractId" />}
         <Todo path="contracts" />
         <Logout path="logout" />
       </Router>
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -25,6 +25,7 @@ import { useStateValue } from './state/useAppState'
 import { HeaderHeading } from './common/components/ExpandableSection'
 import Loading from './common/components/Loading'
 import InspectionDatePage from './page/InspectionDatePage'
+import { requireAdminUser } from './util/userRoles'
 
 const Todo: React.FC<RouteComponentProps> = () => {
   return (
@@ -69,6 +70,7 @@ const Logout: React.FC<RouteComponentProps> = () => {
 
 const App: React.FC = observer(() => {
   const [authState, loading] = useAuth()
+  const [user] = useStateValue('user')
 
   // Listen for the browser close event. Conditionally prompt user when needed.
   let [unsavedFormIds] = useStateValue('unsavedFormIds')
@@ -128,7 +130,7 @@ const App: React.FC = observer(() => {
           path="post-inspection/reports"
           inspectionType={InspectionType.Post}
         />
-        <InspectionDatePage path="inspection-date" />
+        {requireAdminUser(user) && <InspectionDatePage path="inspection-date" />}
         <UserPage path="user" />
         <OperatorContractsListPage path="contract" />
         <EditContractPage path="contract/:contractId" />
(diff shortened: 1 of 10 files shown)


src/inspection/inspectionDate/InsectionDateList.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { LoadingDisplay } from '../../common/components/Loading'
+
+const Header = styled.div`
+  font-size: 1.2rem;
+  margin-bottom: 1rem;
+`
+
+const ListItem = styled.div`
+  padding: 0.75rem;
+  border-bottom: 1px solid #e7e7e7;
```
Divinare, line 15: I guess this color should be from index.css

src/App.tsx
```diff
@@ -127,7 +130,7 @@ const App: React.FC = observer(() => {
           path="post-inspection/reports"
           inspectionType={InspectionType.Post}
         />
-
+        {requireAdminUser(user) && <InspectionDatePage path="inspection-date" />}
```
danieldunderfelt, line 133: Good. This could be done for other pages in the menu as well, for example, the Contracts page should only be accessible by admins.

src/inspection/inspectionDate/InsectionDateList.tsx
danieldunderfelt: Yes! Like this: `var(--lighter-grey)`

src/inspection/inspectionDate/InspectionDateForm.tsx
```diff
@@ -0,0 +1,95 @@
...
+`
+
+const inspectionDateLabels = {
+  startDate: text('start_date'),
+  endDate: text('end_date'),
+}
+
+interface PropTypes {
+  closeInspectionDateList: Function
+  refetchInspectionDateList: Function
```
danieldunderfelt, line 28: Instead of using the Function type, define the function signature here. Eg: `(argName: type) => returnType`.

You can use `unknown` as the return type if the component doesn't care about what the function returns.

src/page/InspectionDatePage.tsx
```diff
@@ -0,0 +1,78 @@
...
+const InspectionDatePage: React.FC<PropTypes> = observer(() => {
+  let loading = false
+  let [isInspectionDateFormVisible, setInspectionDateFormVisibility] = useState<boolean>(false)
+
+  let {
+    data: inspectionDatesQueryResult,
+    loading: areInspectionDatesLoading,
+    refetch: refetchInspectionDates,
+  } = useQueryData(allInspectionDatesQuery, {
+    variables: {},
```
danieldunderfelt, line 40: It is not necessary to set an empty object here.

danieldunderfelt (changes requested):
Very good, just a few observations!

src/App.tsx
Divinare: Yes! I can do it here

Divinare:
Review fixes done ^

src/App.tsx
Divinare: Implemented it here. This error should not happen, frontend should prevent user from selecting invalid date. But of course, BE needs to validate it +1

danieldunderfelt (approved):

PR #17, opened 2020-12-11
User role getters refactor

Divinare:
Relates to: https://github.com/HSLdevcom/bultti/issues/96

### DIFF
--- a/src/inspection/InspectionActions.tsx
+++ b/src/inspection/InspectionActions.tsx
@@ -56,6 +56,10 @@ const InspectionActions = observer(
   ({ inspection, onRefresh, className, style, disabledActions = [] }: PropTypes) => {
     var [season, setSeason] = useStateValue<Season>('globalSeason')
     var [submitActive, setSubmitActive] = useState(false)
+    let hasAdminAccessRights = useHasAdminAccessRights()
+    let hasOperatorUserAccessRights = useHasOperatorUserAccessRights(
+      inspection?.operatorId || undefined
+    )
 
     var isEditing = useMatch(`/:inspectionType/edit/:inspectionId/*`)
 
@@ -173,7 +177,7 @@ const InspectionActions = observer(
     }, [onRefresh, inspection, disabledActions])
 
     let userCanPublish =
-      inspection.status === InspectionStatus.InReview && useHasAdminAccessRights()
+      inspection.status === InspectionStatus.InReview && hasAdminAccessRights
 
     // Pre-inspection which are drafts and post-inspections which are ready can be submitted for approval.
     let inspectionCanBeSubmitted =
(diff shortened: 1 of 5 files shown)


src/inspection/InspectionActions.tsx
```diff
@@ -174,7 +173,7 @@ const InspectionActions = observer(
     }, [onRefresh, inspection, disabledActions])
 
     let userCanPublish =
-      inspection.status === InspectionStatus.InReview && requireAdminUser(user)
+      inspection.status === InspectionStatus.InReview && useHasAdminAccessRights()
```
danieldunderfelt, line 176: This is not valid React hooks usage as the hook will not be called if the first condition returns true. Instead, put the return value from the hook into a variable and use the variable here in the condition.

src/inspection/InspectionActions.tsx
```diff
@@ -227,7 +226,7 @@ const InspectionActions = observer(
           )}
 
           {inspectionCanBeReady &&
-            requireOperatorUser(user, inspection?.operatorId || undefined) &&
+            useHasOperatorUserAccessRights(inspection?.operatorId || undefined) &&
```
danieldunderfelt, line 229: Not valid React hook usage (see the previous comment).

src/inspection/InspectionActions.tsx
```diff
@@ -241,7 +240,7 @@ const InspectionActions = observer(
           {[InspectionStatus.Draft, InspectionStatus.InProduction].includes(
             inspection.status
           ) &&
-            requireAdminUser(user) && (
+            useHasAdminAccessRights() && (
```
danieldunderfelt, line 243: Not valid React hook usage (see the previous comment).

src/inspection/InspectionActions.tsx
```diff
@@ -278,7 +277,7 @@ const InspectionActions = observer(
         </ButtonRow>
         {submitActive &&
           inspectionCanBeSubmitted &&
-          requireOperatorUser(user, inspection?.operatorId || undefined) &&
+          useHasOperatorUserAccessRights(inspection?.operatorId || undefined) &&
```
danieldunderfelt, line 280: Not valid React hook usage (see the previous comment).

src/page/OperatorContractsListPage.tsx
```diff
@@ -115,7 +114,7 @@ const OperatorContractsListPage = observer(({ children }: PropTypes) => {
           <MessageView>Ei sopimusehtoja.</MessageView>
         </MessageContainer>
       )}
-      {!!operator && requireAdminUser(user) && (
+      {!!operator && useHasAdminAccessRights() && (
```
danieldunderfelt, line 117: Not valid React hook usage (see the previous comment).

danieldunderfelt (changes requested):
I like the hooks, they are a good idea, but there are issues with how they are used in the code.

Divinare:
Review fixes done. Now hooks are only called at the top level and not in conditions.

Divinare:
Merging, the changes should be ok

PR #19, opened 2020-12-14
101 procurement unit issues

danieldunderfelt:
Related BE PR: HSLdevcom/bultti#108

Fix UI issues related to findings listed in HSLdevcom/bultti#101.


src/equipmentCatalogue/EditEquipmentCatalogue.tsx
```diff
@@ -0,0 +1,173 @@
...
+          <FlexRow
+            style={{
+              backgroundColor: 'var(--lightest-blue)',
+              margin: '1rem -1rem -1rem',
+              padding: '1rem',
+              borderBottomLeftRadius: '0.5rem',
+              borderBottomRightRadius: '0.5rem',
+              borderTop: '1px solid var(--lighter-grey)',
+            }}>
+            <Button onClick={addDraftCatalogue}>Luo uusi kalustoluettelo</Button>
```
Divinare, line 165: Change text to: "Luo uusi (tyhjä) kalusteluettelo"?

src/equipmentCatalogue/EditEquipmentCatalogue.tsx
```diff
@@ -0,0 +1,173 @@
...
+      )
+    }, [])
+
+    let isDirty = !isEqual(
+      { startDate: catalogue?.startDate, endDate: catalogue?.endDate },
+      pendingCatalogue
+    )
+
+    return (
+      <EditEquipmentCatalogueView>
```
Divinare, line 127: A personal preference would be to make a convention of using names <EntityName>Form for each component that uses ItemForm. So here it would be `EquipmentCatalogueForm`

Divinare: Also another convention I would like:

Have the component export name always the same as the root of our component

For example, now we have in the bottom:
```export default EditEquipmentCatalogue```

we would wrap our component also with the same name ```<EditEquipmentCatalogue></EditEquipmentCatalogue>```

These are little things yes, but following the rule doesn't slow you down. In the long run it slows down when it repeats a lot.

**Review by Divinare — CHANGES_REQUESTED** [review]
Review done!

Testing:
- HelBiltä puuttuu kohteet 118 ja 120 kokonaan
^ these were found

- Kohteeseen 241 ei saa lisättyä kalustoluetteloa. Olen monta kertaa yrittänyt ladata sitä, mutta se ei vaan näy siellä.
^ this was possible

- Kohteessa 139 väärä prosenttiosuus, Bultti antoi yhteensä luvuksi 100,1 %
^ for me it says 100% now

- Kohteessa 259 väärät päästöluokka%osuudet. Kohteeseen on tarjottu 11 A2 Euro 6 energiatehokkaita busseja, mutta - Bultti laskee 3 bussia niistä jollekin toiselle kalustoluokalle. Kalustorekisteri on noiden autojen osalta oikein eli autot ovat Euro 6 energiatehokkaita. Kuva alla:
^ couldn't reprocude this (not that experienced in Bultti yet)

**Comment by Divinare** [comment]
Is there a card for removing exisitng kalusteluettelo?

**Comment by Divinare** [comment]
Also I know you don't like subfolders, but now when it is clear that EquipmentCatalogue is always under ProcurementUnit, I would like to put our <EditEquipmentCatalogue /> under procurementUnit/equipmentCatalogue/ folder. Have src/ received for those views available from sidebar only for now

This follows well the "package-by-feature" rule that is nicely mainly followed in the project that I like

File: src/procurementUnit/ProcurementUnitItem.tsx [hunk 3]
```diff
@@ -288,15 +292,49 @@ const ProcurementUnitItemContent = observer(
...
+              {catalogues.length === 1 ? (
+                <EquipmentCatalogue
+                  startDate={inspectionStartDate}
+                  procurementUnit={procurementUnit}
+                  catalogue={catalogues[0]}
+                  operatorId={procurementUnit.operatorId}
+                  onCatalogueChanged={updateUnit}
+                  editable={catalogueEditable}
+                />
+              ) : (
```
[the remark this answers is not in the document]
danieldunderfelt: You make a good point! I'll change it.

src/procurementUnit/ProcurementUnitItem.tsx
```diff
@@ -288,15 +292,49 @@ const ProcurementUnitItemContent = observer(
...
+                        editable={catalogueEditable}
+                      />
+                    </ExpandableSection>
+                  )
+                })
+              )}
+              {catalogues.length === 0 && (
+                <MessageView>Kilpailukohteella ei ole kalustoluetteloa.</MessageView>
+              )}
+              <EditEquipmentCatalogue
```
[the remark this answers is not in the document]
danieldunderfelt: I'm not sure what you mean. This is the component for editing a catalogue which I extracted from the EquipmentCatalogue component where this functionality was previously. The whole point being that we can use the same EditEquipmentCatalogue component for creating a new catalogue and editing an existing one.

Why would duplicating a part of it be a good solution?

src/equipmentCatalogue/EditEquipmentCatalogue.tsx
```diff
@@ -0,0 +1,173 @@
...
+  catalogue?: EquipmentCatalogue
+  procurementUnit: ProcurementUnit
+  onChange: () => unknown
+}
+
+const EditEquipmentCatalogue = observer(
+  ({ catalogue, procurementUnit, onChange }: PropTypes) => {
+    const catalogueEditMode = useRef<CatalogueEditMode>(
+      !catalogue ? CatalogueEditMode.CREATE : CatalogueEditMode.UPDATE
+    )
```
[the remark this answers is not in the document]
danieldunderfelt: But this is the component that edits a catalogue. It has the pending state, the mutations, and the form. I don't see a good reason why it wouldn't be used for creating a new catalogue.

If we were to do as you propose, we would need multiple pendingCatalogue states all over the place.

As for the isNew prop, I don't think it's needed. If the component gets a catalogue, it edits it. If it doesn't it creates a new one. The CatalogueEditMode enum makes it very clear.

src/equipmentCatalogue/EditEquipmentCatalogue.tsx
danieldunderfelt: I see forms as an implementation detail, they are not important. The task of this component is to _edit_ equipment catalogues, so that's what it is named. The form component itself is named Form because its job is to be a form.

I don't know what you mean by "wrap our component with the same name". The component is EditEquipmentCatalogue and the styled component that wraps it is EditEquipmentCatalogueView. They cannot have the same name so what you suggest would result in the component being wrapped in itself at best or errors at worst.

src/equipmentCatalogue/EditEquipmentCatalogue.tsx
```diff
@@ -0,0 +1,173 @@
...
+              onChange={onChangeCatalogue}
+              onDone={onSaveEquipmentCatalogue}
+              onCancel={onCancelPendingEquipmentCatalogue}
+              isDirty={isDirty}
+              keyFromItem={(item) => item.id}
+              renderInput={renderCatalogueInput}
+              doneLabel={
+                catalogueEditMode.current === CatalogueEditMode.UPDATE
+                  ? 'Tallenna'
+                  : 'Lisää kalustoluettelo'
```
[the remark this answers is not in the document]
danieldunderfelt: Yes, true. I'll update with localized strings.

src/equipmentCatalogue/EditEquipmentCatalogue.tsx
danieldunderfelt: Sure, but the current label already implies that it is empty. Or do you think it implies creating a new catalogue already full with equipment?

danieldunderfelt:
> Also I know you don't like subfolders, but now when it is clear that EquipmentCatalogue is always under ProcurementUnit, I would like to put our under procurementUnit/equipmentCatalogue/ folder. Have src/ received for those views available from sidebar only for now
> 
> This follows well the "package-by-feature" rule that is nicely mainly followed in the project that I like

Equipment catalogues are a distinct feature, so I would keep it as it is. Code structure and UI does not need to be connected, so what appears in the sidebar or what appears under what doesn't have anything to do with how the code is structured.

danieldunderfelt:
> Is there a card for removing exisitng kalusteluettelo?

Yes, here: https://github.com/HSLdevcom/bultti/issues/98

src/equipmentCatalogue/EditEquipmentCatalogue.tsx
Divinare: Maybe no need to add (tyhjä) but I wish we could somehow indicate which kalusteluettelo's are empty. That is an issue right, the operator has to remember to add missing vehicles. Maybe add some light red border around those kalusteluettelo's that are empty?

src/procurementUnit/ProcurementUnitItem.tsx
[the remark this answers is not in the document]
Divinare: Hmm this actually makes sense now when I look at it sorry. Realized you have a form <EditEquipmentCatalogue /> in two places (btw I would use EquipmentCatalogueForm as the name).

src/equipmentCatalogue/EditEquipmentCatalogue.tsx
Divinare: Yeah I guess it's just a personal preference whether to use Edit or Form in the name.

I realized that too, you can't indeed wrap with the same name sadly. Maybe could use as a convention to put "view" in the end to indicate that this is the element that wraps the component

Divinare:
Review fixes are ok!

This can be reviewed after resolving conflicts.

If you want, you can add that red border around empty kalusteluettelo that I suggested above and then merge this.

src/equipmentCatalogue/EditEquipmentCatalogue.tsx
danieldunderfelt: Yeah, it is basically a convention. I have a code template for new components that creates a styled-component with the same name + View in the file. So the idea is that when you have a component and a styled-component that SHOULD have the same name, you add View to the styled-component to differentiate.

PR #20, opened 2020-12-16
HSLdevcom/bultti#88 Remove and refactor redundant old-states from forms.

danieldunderfelt:
I noticed some forms have "old" states for input values, which were used to determine if the form is dirty. Since the "old" state is already available in each component as the main entity that the component is rendering, a separate old state is not needed.

Generally, state should be used sparingly and only when necessary.


Divinare:
This looks fine! The only thing I would still change is our practice of using "pendingProcurementUnit" as procurementUnitInput (change type ProcurementUnitEditInput -> ProcurementUnitInput also). However this should be changed in other places as well

Divinare:
Maybe that is not that big of a deal, so review is ok

PR #23, opened 2020-12-17
105 fix procurement unit issues 2

Divinare:
Relates to: https://github.com/HSLdevcom/bultti/issues/105

1. Procurement units are not listed before the contract is saved

* Showing "Ei ladattua tiedostoa" text when creating new

2.Editing procurement unit will make the unit's contract info disappear from the expandable box header

To fix this, startDate and endDate needed to be added to the single procurementQuery.

A better solution could be to remove currentContracts and make a resolver in ContractResolver to get them. I think that is out of the scope this card though, card -> ?

Btw, I don't really like that startDate and endDate are transfered through the whole component tree, that's ugly. What if we used a store / react context for each page so that you could set a variable there to be accessed from anywhere in the same page?
Card -> ?

3. Remove contract not working

Now you can remove:
- an existing contract
- a new unsaved contract
..and it goes to the /contract page

Btw, I left a TODO there as you can see
fetchPolicy: 'no-cache', // TODO: find a better way than this 
If you have an idea how to do it without using this fetchPolicy, I'm happy to fix it

Also note: Added a small convention to the CONTRIBUTING file

### DIFF
--- a/src/contract/ContractEditor.tsx
+++ b/src/contract/ContractEditor.tsx
@@ -1,7 +1,7 @@
 import React, { useCallback, useEffect, useMemo, useState } from 'react'
 import styled from 'styled-components'
 import { observer } from 'mobx-react-lite'
-import { Contract, ContractInput } from '../schema-types'
+import { Contract as ContractType, ContractInput } from '../schema-types'
 import ItemForm, { FieldLabel, FieldValueDisplay } from '../common/input/ItemForm'
 import { useMutationData } from '../util/useMutationData'
 import {
@@ -32,6 +32,7 @@ import { useUploader } from '../util/useUploader'
 import { SubHeading } from '../common/components/Typography'
 import Table from '../common/components/Table'
 import { text, Text } from '../util/translate'
+import { navigateWithQueryString } from '../util/urlValue'
 
 const ContractEditorView = styled.div`
   padding: 0 1rem;
@@ -55,14 +56,14 @@ const ExpandableFormSectionHeading = styled(HeaderBoldHeading)`
 `
 
 export type PropTypes = {
-  contract: Contract
+  contract: ContractType
   onReset: () => unknown
   onRefresh: () => unknown
   isNew?: boolean
   editable: boolean
 }
 
-function createContractInput(contract: Contract): ContractInput {
+function createContractInput(contract: ContractType): ContractInput {
   return {
     id: contract.id,
     description: contract.description ? contract.description : '',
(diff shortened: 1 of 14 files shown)


src/contract/ContractEditor.tsx
```diff
@@ -369,19 +370,27 @@ const ContractEditor = observer(
...
         let result = await removeContract({
           variables: {
             contractId: contract.id,
           },
         })
 
-        if (pickGraphqlData(result.data)) {
-          onRefresh()
-        }
+        if (result.errors && result.errors.length > 0) return
```
danieldunderfelt, line 390: Code style: always use brackets for if statements, and determine if an array has items with `!== 0`, not `> 0`.

src/contract/ContractEditor.tsx
```diff
@@ -369,19 +370,27 @@ const ContractEditor = observer(
       onReset()
     }, [contract, onReset])
 
-    let [removeContract, { loading: removeLoading }] = useMutationData(removeContractMutation)
+    let [removeContract, { loading: removeLoading }] = useMutationData<ContractType>(
+      removeContractMutation
+    )
```
danieldunderfelt, line 375:
```suggestion
    let [removeContract, { loading: removeLoading }] = useMutationData<ContractType>(
      removeContractMutation,
      {
        refetchQueries: [
          {
            query: contractsQuery,
            variables: {
              operatorId: contract.operatorId,
            },
          },
        ],
      }
    )
```

You can use refetchQueries to automatically refetch some queries after a mutation. This way you do not need to use `no-cache`.

src/page/ContractPage.tsx
```diff
@@ -86,6 +86,7 @@ const OperatorContractsListPage = observer(({ children }: PropTypes) => {
     variables: {
       operatorId: operator?.id,
     },
+    fetchPolicy: 'no-cache', // TODO: find a better way than this (this way contracts get refetched after removing one)
```
danieldunderfelt, line 89: See suggestion about refetchQueries above. No-cache is NEVER a solution. Just forget it exists!

danieldunderfelt (changes requested):
Looks good, but see my comments about no-cache and refetchQueries.

danieldunderfelt:
I agree about having to provide the dates through multiple components. But a store is not the solution, I think React.Context is much better suited for the task. Let's talk about it in Slack.

src/contract/ContractEditor.tsx
Divinare: Sure! I added task to start using ESLint, following this rule could be then automatically ensured: https://github.com/HSLdevcom/bultti/issues/114

src/page/ContractPage.tsx
Divinare: Yup. Need to learn more of how to rely on Apollo's cache

danieldunderfelt: It's a bit weird yes... But it usually has a solution for everything, you just need to dig a bit.

danieldunderfelt (approved):

PR #24, opened 2020-12-18
Added some documentation including DoR and DoD

Divinare:
What do you think?

If you have more "soft rules" of best practices to add to our ideal world, please add them!

I think I would be happy if we followed all of these rules


danieldunderfelt (approved):
Yes, looks good!

PR #39, opened 2021-01-21
151 additions

danieldunderfelt:
Updates to translations and addition of the median age requirement to procurement unit headers. Issue: HSLdevcom/bultti#151


src/dev/PreInspectionDevTools.tsx
```diff
@@ -32,7 +32,16 @@ const InspectionDevTools: React.FC<PropTypes> = observer(({ inspection, onUpdate
 
   return (
     <InspectionDevToolsView>
-      <Button onClick={() => generateEquipment()}>
+      <Button
+        onClick={() => {
+          if (
+            confirm(
+              'Tämä poistaa kaikki tähän tarkastukseen kuuluvia kalustoluetteloiden ajoneuvoja ja generoi testidataa tilalle. Haluatko varmasti jatkaa?'
```
Divinare, line 39: PR is ok (except this needs localization)

danieldunderfelt: This is a dev tool only, I don't think we need to localize it.

Divinare: Nvm, it's just for dev (no need to localize)

PR #40, opened 2021-01-21
WIP: 95 fix date fields into dropdown

Divinare:
Relates to: https://github.com/HSLdevcom/bultti/issues/95

BE PR: https://github.com/HSLdevcom/bultti/pull/169

When creating post inspection results, it's now done by filtering the date options. Other way would be to add possibility to the dropdown to make an item unselectable but maybe leave it for later (make a card if someone proposes it later). Here it would help if the props to our dropdown were typed e.g. IDropdownItem: { key, label, isDisabled }

~~Hmm, I wish we could make InputLabel take hint text as props. Then u could pass hint to a  dropdown that passes it to InputLabel. Make a new card of making <InputLabel /> use <UserHint /> (that wasn't quite simple so I didn't do it here)~~
  - ~~and when this is done, one could add a hint next to dropdown's label if creating pre inspection: "Valittu tarkastusjakso replikoidaan koko viikon ajalle. Valitse aika, jolloin oletat olevan vähiten erikoispäivätyypin liikennettä"~~
  ->**Card**: https://github.com/HSLdevcom/bultti/issues/174

Minor proposals (TODO in this card?)
- ~~should bulttiDate() (that was created in this PR) be used instead of format(date, DATE_FORMAT) everywhere?~~
- ~~- rename readableDate() -> getReadableDate()?~~
- ~~-Fix in a new card: Departure => PlannedDeparture - Create a migraton file that renames departure as planned_departure in db~~- -> Card: https://github.com/HSLdevcom/bultti/issues/171

### DIFF
--- a/src/common/input/Dropdown.tsx
+++ b/src/common/input/Dropdown.tsx
@@ -34,13 +34,11 @@ const SelectButton = styled(Button).attrs({ size: ButtonSize.MEDIUM })<{
   font-size: 1rem;
   justify-content: flex-start;
   display: flex;
-
   svg {
     display: ${(p) => (p.disabled ? 'none' : 'block')};
     margin-left: auto;
     margin-right: 0;
   }
-
   ${(p) =>
     !p.disabled
       ? css`
@@ -48,7 +46,6 @@ const SelectButton = styled(Button).attrs({ size: ButtonSize.MEDIUM })<{
             background: ${(p) => (p.theme === 'light' ? '#fafafa' : 'var(--lighter-grey)')};
             color: var(--dark-grey);
             border-color: var(--blue);
-
             svg * {
               fill: ${(p) => (p.theme === 'light' ? 'var(--blue)' : 'var(--dark-grey)')};
             }
@@ -86,11 +83,11 @@ const DropdownItem = styled.li<{ highlighted: boolean }>`
 export type DropdownProps = {
   disabled?: boolean
   label?: string
-  items: any[]
+  items: any[] // any object (remember to pass itemToString, itemToLabel), array, { field, value } object
   onSelect: (selectedItem: any | null) => unknown
-  itemToString?: string | ((item: any | null) => string)
-  itemToLabel?: string | ((item: any | null) => string)
-  selectedItem?: any
+  itemToString?: string | ((item: any | null) => string) // property of given object to get value from
+  itemToLabel?: string | ((item: any | null) => string) // property of given object to get label from
+  selectedItem?: any // TODO: add documentation of this property or change any
   className?: string
   style?: CSSProperties
   theme?: ThemeTypes
@@ -174,6 +171,7 @@ const Dropdown: React.FC<DropdownProps> = observer(
               <>
                 {items.map((item, index) => (
                   <DropdownItem
+                    key={`dropdown-item-${index}`}
                     theme={theme}
                     highlighted={highlightedIndex === index}
                     {...getItemProps({
--- a/src/common/input/Dropdown.tsx
+++ b/src/common/input/Dropdown.tsx
@@ -83,10 +83,10 @@ const DropdownItem = styled.li<{ highlighted: boolean }>`
 export type DropdownProps = {
   disabled?: boolean
   label?: string
-  items: any[] // any object (remember to pass itemToString, itemToLabel), array, { field, value } object
+  items: any[] // any object (remember to pass itemToString, itemToLabel), array, { field, value } object.
   onSelect: (selectedItem: any | null) => unknown
-  itemToString?: string | ((item: any | null) => string) // property of given object to get value from
-  itemToLabel?: string | ((item: any | null) => string) // property of given object to get label from
+  itemToString?: string | ((item: any | null) => string) // property of given object to get value from or a function that returns the value.
+  itemToLabel?: string | ((item: any | null) => string) // property of given object to get label from or a function that returns the label.
   selectedItem?: any // TODO: add documentation of this property or change any
   className?: string
   style?: CSSProperties
--- a/src/executionRequirement/PostInspectionExecutionRequirements.tsx
+++ b/src/executionRequirement/PostInspectionExecutionRequirements.tsx
@@ -24,7 +24,7 @@ import {
   updateObservedExecutionRequirementValuesMutation,
 } from './executionRequirementsQueries'
 import { groupBy, toString } from 'lodash'
-import { readableDateRange } from '../util/formatDate'
+import { getReadableDateRange } from '../util/formatDate'
 import { FlexRow } from '../common/components/common'
 import { round } from '../util/round'
 import { getTotal } from '../util/getTotal'
@@ -190,7 +190,7 @@ const PostInspectionExecutionRequirements = observer(({ isEditable }: PropTypes)
         areaName,
         Object.entries<ObservedExecutionRequirement[]>(
           groupBy<ObservedExecutionRequirement>(areaReqs, (req) =>
-            readableDateRange({ start: req.startDate, end: req.endDate })
+            getReadableDateRange({ start: req.startDate, end: req.endDate })
           )
         ),
       ]),
--- a/src/executionRequirement/PostInspectionExecutionRequirements.tsx
+++ b/src/executionRequirement/PostInspectionExecutionRequirements.tsx
@@ -180,9 +180,10 @@ const PostInspectionExecutionRequirements = observer(({ isEditable }: PropTypes)
     }
   }, [inspection, removeRequirements, observedRequirements, isEditable])
 
-  let requirementsByAreaAndWeek: Array<
-    [string, Array<[string, ObservedExecutionRequirement[]]>]
-  > = useMemo(
+  let requirementsByAreaAndWeek: Array<[
+    string,
+    Array<[string, ObservedExecutionRequirement[]]>
+  ]> = useMemo(
     () =>
       Object.entries<ObservedExecutionRequirement[]>(
         groupBy<ObservedExecutionRequirement>(observedRequirements, 'area.name')
@@ -190,7 +191,7 @@ const PostInspectionExecutionRequirements = observer(({ isEditable }: PropTypes)
         areaName,
         Object.entries<ObservedExecutionRequirement[]>(
           groupBy<ObservedExecutionRequirement>(areaReqs, (req) =>
-            readableDateRange(req.startDate, req.endDate)
+            readableDateRange({ startDate: req.startDate, endDate: req.endDate })
           )
         ),
       ]),
--- a/src/executionRequirement/PostInspectionExecutionRequirements.tsx
+++ b/src/executionRequirement/PostInspectionExecutionRequirements.tsx
@@ -180,10 +180,9 @@ const PostInspectionExecutionRequirements = observer(({ isEditable }: PropTypes)
     }
   }, [inspection, removeRequirements, observedRequirements, isEditable])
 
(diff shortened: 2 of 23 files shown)


src/inspection/inspectionSelectDates.tsx
```diff
@@ -0,0 +1,140 @@
...
+    let selectedItem: DateOption | null =
+      inspectionInput.inspectionStartDate && inspectionInput.inspectionEndDate
+        ? {
+            label: readableDateRange({
+              startDate: inspectionInput.inspectionStartDate,
+              endDate: inspectionInput.inspectionEndDate,
+            }),
+            value: {
+              startDate: inspectionInput.inspectionStartDate,
+              endDate: inspectionInput.inspectionEndDate!,
```
danieldunderfelt, line 71: Use date-fns Interval interface here too. OK to use strings as values though.

src/inspection/inspectionSelectDates.tsx
```diff
@@ -0,0 +1,140 @@
...
+            onSelect={onSelectDates}
+            selectedItem={selectedItem}
+          />
+        )}
+      </InspectionSelectDatesView>
+    )
+  }
+)
+
+const _getPreInspectionDateOptions = (): DateOption[] => {
```
danieldunderfelt, line 92: Standalone functions should be defined with the `function` keyword.

And drop the underscore, I don't like it, except for utility properties on objects.

src/inspection/inspectionSelectDates.tsx
```diff
@@ -0,0 +1,140 @@
...
+  let startDate = new Date()
+  let endDate = new Date(startDate)
+  endDate.setDate(endDate.getDate() + 90)
```
danieldunderfelt, line 95: Javascript's Date object is horrible and we should use date-fns functions wherever possible. I have no idea what's going on here. Suggestion:

```suggestion
  let endDate = addSeconds(startDate, 90)
```

Substitute "addSeconds" with the function that corresponds to what unit the "90" is supposed to be.

src/inspection/inspectionSelectDates.tsx
```diff
@@ -0,0 +1,140 @@
...
+  let dateOptionsEndDates = eachWeekOfInterval({
+    start: startDate,
+    end: endDate,
+  })
```
danieldunderfelt, line 99: This Is The Way.

src/inspection/inspectionSelectDates.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+  let dateOptions: DateOption[] = dateOptionsEndDates.map((endDate) => {
+    let startDate = startOfWeek(endDate, { weekStartsOn: 1 })
+    let value = {
+      startDate,
+      endDate,
+    }
+    let label = readableDateRange({ startDate, endDate })
```
danieldunderfelt, line 107: Date-fns Interval interface for all of this too!

src/inspection/inspectionSelectDates.tsx
```diff
@@ -0,0 +1,140 @@
...
+      label,
+      value,
+    }
+  })
+  return dateOptions
+}
+
+const _getPostInspectionDateOptions = (
+  inspectionDatesQueryResult: InspectionDate[]
+): DateOption[] => {
```
danieldunderfelt, line 118: Same thing as above, use `function` and drop the underscore.

src/inspection/inspectionSelectDates.tsx
```diff
@@ -0,0 +1,140 @@
...
+  let dateOneMonthAgo = sub(new Date(), { months: 1 })
```
danieldunderfelt, line 119: I'd use `subMonths()` but this is not really wrong either.

src/util/formatDate.ts
```diff
@@ -33,3 +39,7 @@ export function readableDateRange(startDate: AcceptedDateFormat, endDate: Accept
     READABLE_DATE_FORMAT
   )}`
 }
+
+export function getBulttiDate(date: Date): string {
+  return format(date, DATE_FORMAT)
+}
```
danieldunderfelt, line 45: Good to encapsulate this in a function, but I wouldn't use "bultti" in the function name. Also use the `AcceptedDateFormat` arg type and `getDateObject` to make the function more flexible.

And yes we should probably replace all `format(dateStr, DATE_FORMAT)` with this function.

danieldunderfelt (changes requested):
Mostly good, but some comments.

A big theme is using the date-fns Interval interface to represent all date ranges, even for functions that don't come from date-fns. Entities (Inspection and InspectionDateRange) should of course keep their current date properties as they are, I'm only talking about how date ranges are handled in the code. The value of `start` and `end` on the ranges should be Date where possible, except in cases where there would be a lot of converting back and forth between Date and string.

src/executionRequirement/PostInspectionExecutionRequirements.tsx
```diff
@@ -180,17 +180,18 @@ const PostInspectionExecutionRequirements = observer(({ isEditable }: PropTypes)
...
+  ]> = useMemo(
     () =>
       Object.entries<ObservedExecutionRequirement[]>(
         groupBy<ObservedExecutionRequirement>(observedRequirements, 'area.name')
       ).map(([areaName, areaReqs]) => [
         areaName,
         Object.entries<ObservedExecutionRequirement[]>(
           groupBy<ObservedExecutionRequirement>(areaReqs, (req) =>
-            readableDateRange(req.startDate, req.endDate)
+            readableDateRange({ startDate: req.startDate, endDate: req.endDate })
```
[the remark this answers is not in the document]
Divinare: I think startDate and endDate are better variable names but changed this anyway if it's a convention

src/common/input/Dropdown.tsx
```diff
@@ -86,11 +83,11 @@ const DropdownItem = styled.li<{ highlighted: boolean }>`
 export type DropdownProps = {
   disabled?: boolean
   label?: string
-  items: any[]
+  items: any[] // any object (remember to pass itemToString, itemToLabel), array, { field, value } object
```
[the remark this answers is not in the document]
Divinare: Yup, typing inspead of any is always nice. But sometimes you need to use `any` like when making a generic update value func from a form. Having much different prop types increases complexity though

src/inspection/inspectionSelectDates.tsx
Divinare: Yeah but I wouldn't use the same variable names just because a good library uses them. I think variable names should specify what `type` they are so you don't have to look around them. But yeah it's not that a big deal here.

Divinare: And yeah I don't mean that is always the case :D

sometimes it's just ok to do this for example:
```
dateOptions.sort((a: DateOption, b: DateOption) => {
      return a.value.startDate.getTime() < b.value.startDate.getTime() ? -1 : 1
})
    ```

File: src/inspection/inspectionSelectDates.tsx (hunk 5 above)
Comment (Divinare) [reply src/inspection/inspectionSelectDates.tsx]: It's enough to say it in 1 place, I will then change it everywhere else :>

File: src/inspection/inspectionSelectDates.tsx (hunk 6 above)
Comment (Divinare) [reply src/inspection/inspectionSelectDates.tsx:118]: I see nothing wrong with the underscore though. It indicates that the func is a private function declared in this file. It's just a helper function, nothing interesting there. Also I put it to the end of the file so that you have less things to look at.

Comment (Divinare) [reply src/inspection/inspectionSelectDates.tsx:118]: I agree on using function though!

File: src/inspection/inspectionSelectDates.tsx (hunk 7 above)
Comment (Divinare) [reply src/inspection/inspectionSelectDates.tsx:119]: Using subMonths(), maybe it's slightly more readable.

**Comment by Divinare** [comment]
Review fixes & suggestions done. There are some more suggestions that aren't dealt with (check PR description above)

File: src/inspection/inspectionSelectDates.tsx [hunk 11]
```diff
@@ -0,0 +1,140 @@
...
+    let [
+      queryInspectionDates,
+      { data: inspectionDatesQueryResult, loading: areInspectionDatesLoading },
+    ] = useLazyQueryData<InspectionDate[]>(allInspectionDatesQuery)
+
+    let _queryPostinspectionDateOptions = async () => {
+      if (!areInspectionDatesLoading && !inspectionDatesQueryResult) {
+        await queryInspectionDates()
+      }
+    }
```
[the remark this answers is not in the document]
danieldunderfelt: I would still prefer this if possible.

src/inspection/inspectionSelectDates.tsx
```diff
@@ -0,0 +1,140 @@
...
+    let dateOptions: DateOption[] = []
+    if (inspectionType === InspectionType.Pre) {
+      dateOptions = _getPreInspectionDateOptions()
+    } else {
+      if (inspectionDatesQueryResult) {
+        dateOptions = _getPostInspectionDateOptions(inspectionDatesQueryResult)
+      } else {
+        _queryPostinspectionDateOptions()
+      }
+    }
```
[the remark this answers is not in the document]
danieldunderfelt: And this!

src/inspection/inspectionSelectDates.tsx
danieldunderfelt: My point is that if there are places in the code where you define a date range as an object, I would prefer if we used the same interface from date-fns. So if you were going to do this:

```
let range = {
  startDate: dateValue,
  endDate: dateValue
}

```

Instead do this:

```
let range = {
  start: dateValue,
  end: dateValue
}

```

That way, every date range defined in the app adheres to the same interface. Of course, if the object is _not_ a date range (eg it has other props too) then startDate and endDate is ok to keep.

src/inspection/inspectionSelectDates.tsx
```diff
@@ -0,0 +1,140 @@
...
+    let label = getReadableDateRange({ start: startDate, end: endDate })
+    return {
+      label,
+      value,
+    }
+  })
+  return dateOptions
+}
+
+function _getPostInspectionDateOptions(
```
danieldunderfelt, line 116:
```suggestion
function getPostInspectionDateOptions(
```

src/inspection/inspectionSelectDates.tsx
```diff
@@ -0,0 +1,140 @@
...
+            onSelect={onSelectDates}
+            selectedItem={selectedItem}
+          />
+        )}
+      </InspectionSelectDatesView>
+    )
+  }
+)
+
+function _getPreInspectionDateOptions(): DateOption[] {
```
danieldunderfelt, line 92:
```suggestion
function getPreInspectionDateOptions(): DateOption[] {
```

danieldunderfelt (changes requested):
A few comments still! But it is looking good.

src/common/input/Dropdown.tsx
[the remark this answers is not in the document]
danieldunderfelt: Yeah let's not overcomplicate it.

src/inspection/inspectionSelectDates.tsx
danieldunderfelt: I just think the underscore is very ugly and unnecessary, I avoid it like the plague. If it's a private helper it is enough to just not export it.

Divinare:
Review fixes done!

Divinare:
Review fixes done again.

danieldunderfelt (approved):
Great work!

PR #42, opened 2021-01-29
90 inspection page sticky header

Divinare:
Relates to: https://github.com/HSLdevcom/bultti/issues/90

Top bar is now sticky and slimmer! I didn't move <InspectionActionsRow /> into the header (next to Päivitä button) yet because I think that needs further planning.

Maybe the timeline could replace those <InspectionActionsRow /> buttons instead. Will have to wait until that task gets planned out.

Hmm, I didn't implement "when form is ok, activate tulokset tab". To implement that, I think I should change how <Tabs> are used from
```
<Tabs>
    <InspectionEditor
    name="create"
    path="/"
    label={text('inspectionPage_inspectionInformation')}
    loading={inspectionLoading}
    refetchData={refetch}
    inspection={inspection}
    />
    <InspectionPreview
    inspection={inspection}
    path="results"
    name="results"
    label={text('inspectionPage_results')}
    />
</Tabs>
```

to

```
<Tabs>
  <Tab isDisabled={<pass_boolean_here}>
    <InspectionEditor
      name="create"
      path="/"
      label={text('inspectionPage_inspectionInformation')}
      loading={inspectionLoading}
      refetchData={refetch}
      inspection={inspection}
    />
  </Tab>
  <Tab>
    <InspectionPreview
      inspection={inspection}
      path="results"
      name="results"
      label={text('inspectionPage_results')}
    />
  </Tab>
</Tabs>
```
right? Make this in a new card so that this one gets moved forward?


danieldunderfelt:
I don't think it is necessary to activate the Results tab automatically, it can be disorienting for the user.

danieldunderfelt (approved):
Code style discussions can be had in Slack. This works nicely, approved!

PR #43, opened 2021-02-02
HSLdevcom/bultti#90 Small inspection editor style updates.

danieldunderfelt:
Additional smaller UI fixes. Issue: HSLdevcom/bultti#90

- Expandable and hideable inspection meta
- Quick and dirty patch for Apollo Client to allow fast-refresh edits in Inspection views
- Minor spacing fixes to inspection page components

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -3,6 +3,7 @@
 - Use `<button>` element when you have an element which the user interacts with through clicking (never use eg. `<div>` unless necessary).
 - Use `{}` for empty objects
 - Use `''` for empty strings
+- Avoid using `any` type as much as you can. Only use it if you have to. For example when using useQueryData hook, give it a return type
 
 ### Typescript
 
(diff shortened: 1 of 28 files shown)


Divinare:
TODO:
* merge user info & inspection info toggle containers as one
* create a PageContainer and start using it in every page

Note: I will continue this PR!

Divinare:
InspectionMeta taken out to reduce duplcated shown info. Might seem to be wrong at first sight because it uses getAllUpdatedByUsers() and <UserRelations> doesn't but actually UserRelations gets all userRelations, sorts them out by updatedAt and then prints the one that is the latest one (i'm not sure though, didn't test this with multiple users... should make a new user to test this)

Well, if we want to return InspectionMeta that can easilly be done. But I think we need to plan how this should look like in another task.

Still TODO:
create a PageContainer and start using it in every page

Divinare:
Actually decided to revert InspectionMeta because it will be easier to demo / discuss of how we want it to look like when we have it in UI :+1:

Divinare:
Now changes can be reviewed. I think this PageContainer was great because now we have harmonized default styles for each page!

The only "downside" was that when using a sticky header, I had to add this to override some PageContainer styles:
```EditInspectionPageContainer```
But that's fine. If we decide to make sticky headers to every page, then overflow-y: auto can be added to PageContainer.

PR #44, opened 2021-02-03
167 remove create inspection buttons

Divinare:
Closes https://github.com/HSLdevcom/bultti/issues/167

Noticed that there was a need for parameterization in localization in the same file. So as a bonus, now text() supports keyValueMap which was very much needed!


danieldunderfelt (approved):
Looks good! LGTM

PR #45, opened 2021-02-04
174 hint param to support input label

Divinare:
Closes https://github.com/HSLdevcom/bultti/issues/174

Now Input, InputLabel and Dropdown supports hintText param. I think we can make other components support hintText too as we go (should be easy since InputLabel now supports it)

Btw I sorted those Input params. Imo we don't need to nitpick param order in PR's but I wish you could try to put them in some specific order. I usually list the core ones first and then "not so core" props. And I also like to put methods lastly. Could even document this to conventions as "not strict rules" or smth

### DIFF
--- a/src/common/components/form.tsx
+++ b/src/common/components/form.tsx
@@ -1,6 +1,9 @@
 import styled from 'styled-components/macro'
 import { Column } from './common'
 import { ThemeTypes } from '../../type/common'
+import { observer } from 'mobx-react-lite'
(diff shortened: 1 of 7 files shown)


danieldunderfelt (approved):
Looks good!

I don't necessarily agree with the importance of ordering props, but I will try to keep it in mind. Putting required props first and optional props after them is a good idea though, then you can quickly scan the props and get an idea of what the component requires.

PR #46, opened 2021-02-04
179 localization harmonization

Divinare:
Closes https://github.com/HSLdevcom/bultti/issues/179

There we go :dancers:, keys are harmonized.

Tried to make some keys better but didn't scan through them all. At least they should be better now

Also removed some keys that weren't used (or added some missing keys) etc.


danieldunderfelt (approved):
Good!

PR #49, opened 2021-02-09
Cypress init

Divinare:
Closes https://github.com/HSLdevcom/bultti/issues/37

BE PR: https://github.com/HSLdevcom/bultti/pull/189

* Initial cypress tests for logging in with different roles. We can continue implementing e2e tests after this card in other cards:
    * initial e2e tests for contract / procurementUnit -> card?
    * initial e2e tests for inspections -> card?
* Note: there is a small issue with .env, let me explain: currently .env is added into the repo (it's ok, since it doesn't have any real secrets). Now the new Idea here is to have .env (.gitignored) and default.env (template) so that new developer can, for local development, just rename default.env -> .env, fill secrets and that's all
    * if this solution is ok, .env needs to be removed from git (currently .gitignore doesn't ignore it since it's already in git...)
EDIT: we can have multiple .env files too, but if so, maybe secrets could be in a single .env file and the rest non-secrets in other? Note: I'd like to have one .env though, don't atm see advantage of having multiple .envs


danieldunderfelt:
Works nicely! I removed the .env from the repo in this branch.

danieldunderfelt (approved):
Looks good! Check BE PR comments before merging.

PR #52, opened 2021-02-16
150 inspection update button fixes

Divinare:
Closes: https://github.com/HSLdevcom/bultti/issues/150

Found 2 bugs and fixed them:
* Bug 1: Pre inspection name startDate / endDate saving didnt work
  * The reason for this was simple. Found from BE inspectionResolver: "Allow only name update if inspection is not in draft state."
  --> Had to just disable editing dates if not in DRAFT mode
* Bug 2: When date saved, isDirty didn't get updated (save button was active even after saving)

Removed update buttons from Inspection pages
  * yes, didn't remove procurementUnit update button because it's being used at procurementUnit page and it does have a job there - instead that update button should be moved to the header in procurementUnit page (should I do it here?)
  
Note: also refactored updateInspectionCb - removed useless code
  * the idea in a save method is that it shouldn't be called when a form is not dirty. So saveCb is just a save call

### DIFF
--- a/src/common/components/Loading.tsx
+++ b/src/common/components/Loading.tsx
@@ -8,7 +8,7 @@ const spin = keyframes`
   from {
     transform: rotate(0deg);
   }
-  
+
   to {
     transform: rotate(360deg);
   }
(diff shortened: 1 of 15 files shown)


danieldunderfelt (approved):
Everything looks good! LGTM

PR #53, opened 2021-02-17
Contract form: static operator input + some refacotring

Divinare:
Closes https://github.com/HSLdevcom/bultti/issues/175

Static operator name implemented. Had to think where to use globalOperator and where to use existingContract operator.

Also did some minor refactoring, mainly renaming and moved the code of creating new contract into the contract form out of it's parent. I think it the code got better now.

Also it crashes here: http://localhost:3001/contract/new?season=KE2021&operator=22
<PageTitleOperator>{globalOperator.operatorName}</PageTitleOperator>
because globalOperator is not set (yet). The EditContactPage gets rendered before code gets to set it. I think we need a task where we would in global level and in one place do globalOperator loading and ONLY after that, render either the main page content. If globalOperator is not set, you would get notification to set it --> Card?
  * BTW, contractPage currently shows all contracts of other operators when globalOperator is not set. Also this should be changed so that even if you have 1 operator set, you would see all operaotr contracts (if you are admin). The contracts of the selected globalOperator could be shown above all the other contracts

Ah one additional note: maybe contract.operator could me made as not optional in backend's Contract entity?


Divinare:
No crash should happen. Also found a nasty bug that allowed you to change contract operator of an existing contract (this wasn't a bug before but it was made by this pr).

src/contract/ContractEditor.tsx
```diff
@@ -53,14 +55,13 @@ const ExpandableFormSectionHeading = styled(HeaderBoldHeading)`
 `
 
 export type PropTypes = {
-  contract: ContractType
-  onReset: () => unknown
+  existingContract: Contract | null
```
danieldunderfelt, line 58: `contract` is fine, the "existing" prefix is redundant.

src/contract/ContractEditor.tsx
```diff
@@ -221,8 +219,34 @@ const renderEditorLabel = (key, val, labels) => {
...
-  ({ contract, onReset, onRefresh, isNew = false, editable }: PropTypes) => {
-    let [pendingContract, setPendingContract] = useState(createContractInput(contract))
+  ({ existingContract, onRefresh, isNew = false, editable }: PropTypes) => {
+    let [globalOperator] = useStateValue<Operator>('globalOperator')
+    let newContract: Partial<Contract> = {
+      operatorId: globalOperator?.id,
+      operator: globalOperator,
+      startDate: getDateString(new Date()),
+      endDate: getDateString(addYears(new Date(), 1)),
+    }
```
danieldunderfelt, line 229: The new contract should be created in the `createContractInput` function. There doesn't seem to be any reason for it to be here.

src/contract/ContractEditor.tsx
```diff
@@ -221,8 +219,34 @@ const renderEditorLabel = (key, val, labels) => {
...
+      [isNew, newContract, existingContract]
+    )
+    let resetChanges = useCallback(() => {
+      setPendingContract(initialContract)
+    }, [initialContract])
+    let [pendingContract, setPendingContract] = useState<ContractInput>(initialContract)
+
+    useEffect(() => {
+      if (isNew && pendingContract.operatorId !== globalOperator.operatorId) {
+        console.log('updatign! ! ')
```
danieldunderfelt, line 242: Leftover?

danieldunderfelt (changes requested):
It works well, but I have some comments.

src/contract/ContractEditor.tsx
Divinare: Yeah... =|

src/contract/ContractEditor.tsx
Divinare: New contractInput is created at createContractInput(), Partial<Contract> has no constructor

src/contract/ContractEditor.tsx
Divinare: Well, you then have to rename contract as existingContract when you read the props in. And existingContract makes clear that it's not a new contract that can be given

danieldunderfelt: I just feel that it is redundant, since providing a contract for this prop already means that it is"existing". This is the _main_ contract for this component, so I don't think it needs any prefixes, even inside the component. If it is not provided, the component makes a newContract.

Also, this should be optional: `contract?: Contract` instead of `Contract | null`.

src/contract/ContractEditor.tsx
danieldunderfelt: Yes, I was thinking that createContractInput could take a Contract OR the globalOperator and create this new partial contract based on it. Now it creates this newContract object every time the component updates which seems unnecessary. Of course it's not a huge performance impact, but stuff like this should not be by itself like this in the render function.

At least put it in the useMemo, inline in the createContractInput call.

Divinare:
Fixes now done. Made a small change to useQueryData so that I could use any typing with it. This card is to finish the job: https://github.com/HSLdevcom/bultti/issues/209

Note: as I did review fixes I found a new bug that is now fixed. As you enter: https://bultti-ui.azurewebsites.net/contract/new?season=KE2021 you see empty page with no msg, that's now fixed!

If you now refresh here:
http://localhost:3001/contract/new?season=KE2021

It will say "Valitse liikennöitsijä."

danieldunderfelt:
> Note: as I did review fixes I found a new bug that is now fixed. As you enter: https://bultti-ui.azurewebsites.net/contract/new?season=KE2021 you see empty page with no msg, that's now fixed!

Super, good catch!

danieldunderfelt (approved):
LGTM

src/contract/ContractEditor.tsx
```diff
@@ -221,8 +219,36 @@ const renderEditorLabel = (key, val, labels) => {
...
+          operatorId: globalOperator?.id,
+          operator: globalOperator,
+          startDate: getDateString(new Date()),
+          endDate: getDateString(addYears(new Date(), 1)),
+        }
+        return createContractInput(newContract)
+      } else {
+        return createContractInput(contract!)
+      }
+    }, [isNew, contract])
```
danieldunderfelt, line 236:
```suggestion
    }, [isNew, contract, globalOperator])
```

PR #58, opened 2021-03-10
206 inspection date range selection fixes

Divinare:
Closes https://github.com/HSLdevcom/bultti/issues/206

* When merged, this card can be removed: https://github.com/HSLdevcom/bultti/issues/224
* I removed check from filtering observed date options "Only dates that are older than 1 month are valid"
   * As now the first inspection date is selected by default, I see no reason to filter here (because invalid value could still end up as being selected because of the default selection). Instead, before prosessing inspections, we should validate that the currently selected inspectionStartDate & inspectionEndDate are valid. In which card should we document this change?

* If date range doesnt exist for a season creating post inspection, an error is thrown. Card to localize this (and other) messages: https://github.com/HSLdevcom/bultti/issues/227

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,6 @@ src/data
 
 # Visual Studio Code
 .vscode
+
+# Webstorm
+.idea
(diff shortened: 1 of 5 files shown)


danieldunderfelt (approved):
Works and looks good! Merging.

PR #59, opened 2021-03-10
203 refactor hfp loading

danieldunderfelt:
Closes https://github.com/HSLdevcom/bultti/issues/203

UI PR for HSLdevcom/bultti#203

Server PR: https://github.com/HSLdevcom/bultti/pull/228

- Add button to clear the cache from the UI
- Remove "Load HFP box from post-inspection editor
- Add HFP status panel to Inspection Dates view


Divinare (changes requested):
Good job, date range loading worked after some struggling with concurrency problems due to wrong .env params. It's better usability now compared to having UI in inspection page!
 
I think we should add some documentation about some .env params "what params one should use for local development"
HFP_TASK_DATE_CONCURRENCY=1
WORKER_CONCURRENCY=1


* I think we need some kind of mechanism to prevent reloading HFP-data
  * Like load HFP-data twice for each day and check with md5sums that the data for the day is the same? Then mark data for that day usable
    * This would require some new db entity like "HFPDateStatus { date: Date, md5sum?: string }
    * Once in sync, manual data load/reload HFP-data via"tarkastusjakson HFP ohjauspaneeli" could be prevented

Testing:

1) Bug: date progress doesnt show

Also tried to see what is going on, pretty hard to know when all variables seem to be undefined when logging :P
hfpTaskProgressData undefined
InspectionDateHfpControl.tsx:203 requestedHfpDateRanges  undefined
InspectionDateHfpControl.tsx:204 hfpStatusData  undefined
InspectionDateHfpControl.tsx:205 dateProgress  Map(0) {}

--> If subscriptions dont work in azure anyway, the quick fix would be to just simply fetch InspectionDate.hfpDataStatus when user presses f5.

danieldunderfelt:
I'll just remove the progress display from the UI for now.

danieldunderfelt:
Review fixes done! By removing a lot of code!

danieldunderfelt:
> I think we should add some documentation about some .env params "what params one should use for local development"
> HFP_TASK_DATE_CONCURRENCY=1
> WORKER_CONCURRENCY=1

Yes we should, but this is also up to the developer. The app works with some amount of concurrency, it depends on the local machine how much it can handle. Also, this concerns the server and not the UI.

> I think we need some kind of mechanism to prevent reloading HFP-data
> Like load HFP-data twice for each day and check with md5sums that the data for the day is the same? Then mark data for that day usable
> This would require some new db entity like "HFPDateStatus { date: Date, md5sum?: string }
> Once in sync, manual data load/reload HFP-data via"tarkastusjakson HFP ohjauspaneeli" could be prevented

This is an interesting idea which we should explore more! But it is also server stuff and this is the UI PR...

Divinare (approved):
LGTM, good that we removed progress - it wasn't that useful anyway I think

PR #60, opened 2021-03-10
208 fix unsaved changes prompt

Divinare:
Closes https://github.com/HSLdevcom/bultti/issues/208

Prompt bugs that are fixed here:

1) Insert a new procurementUnit list -> prompts if leaving page
2) Go and create a new contract -> save -> prompts if leaving page
3) Inspection name change -> doesn't prompt if unsaved changes

(didn't find any other prompt issuses while testing - now everything seem to work!)

NOTE: when creating new inspection, it prompts about unsaved changes if trying to leave. That is due to name being dirty when it shouldnt be. It will be fixed in my date dropdown PR! So dont mind about that

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,6 @@ src/data
 
 # Visual Studio Code
 .vscode
+
+# Webstorm
+.idea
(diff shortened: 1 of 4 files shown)


danieldunderfelt:
Inspection date page also prompts after creating a new InspectionDate.

danieldunderfelt (approved):
Couldn't find any prompts anymore! LGTM

PR #61, opened 2021-03-12
List contracts of a procurement unit

Divinare:
Closes https://github.com/HSLdevcom/bultti/issues/116

Hmm, should LinkButton be used in Contracts index page e.g. http://localhost:3001/contract?season=KE2021&operator=22

Should button.tsx be moved to /buttons as well?

Things I didn't do yet because I'm not sure if we both agree on this. Also I would split components from button.tsx into their own components under /buttons, but if we dont agree on this -> can just move buttons.tsx under /buttons

Date used in comparison:
* in inspection page: inspection.startDate
* in procurementUnitList: season.startDate


src/procurementUnit/ProcurementUnitItem.tsx
```diff
@@ -75,6 +78,24 @@ const ProcurementUnitItem: React.FC<PropTypes> = observer(
...
+        .filter((contract: Contract) => {
+          return isWithinInterval(new Date(), {
+            start: getDateObject(contract.startDate),
+            end: getDateObject(contract.endDate),
+          })
+        })
+        .sort((a: Contract, b: Contract) => {
+          return getDateObject(a.startDate).getTime() < getDateObject(b.startDate).getTime()
+            ? 1
+            : -1
```
danieldunderfelt, line 96: This is certainly not *wrong*, but I'd just use orderBy from lodash. `sort` is pretty low-level.

src/procurementUnit/ProcurementUnitItem.tsx
```diff
@@ -75,6 +78,24 @@ const ProcurementUnitItem: React.FC<PropTypes> = observer(
       ? procurementUnit?.area?.name
       : OperatingAreaName.Unknown
 
+    // The logic behind finding displayed contractUnit: selecting the latest contract that is currently valid
+    // meaning if there are multiple contracts being valid, we select the one with the latest startDate
+    let displayedContractUnit: Contract | null = null
+    if (currentContracts && currentContracts.length > 0) {
+      displayedContractUnit = currentContracts
```
danieldunderfelt, line 85: You could probably just do:

```suggestion
    let displayedContractUnit = currentContracts
```

And have it be undefined (instead of null) if no contract was found. Same outcome.

danieldunderfelt (approved):
Looks good! We can use a subfolder for buttons if you want. Left a few comments.

src/procurementUnit/ProcurementUnitItem.tsx
Divinare: sort seems to be faster at least according to this test: https://www.measurethat.net/Benchmarks/Show/7726/0/lodash-orderby-vs-arraysort

I'll stick with sort

src/procurementUnit/ProcurementUnitItem.tsx
Divinare: Using Conditional ? operator now

Divinare:
Review fixes done!

src/procurementUnit/ProcurementUnitItem.tsx
danieldunderfelt: Just remember that we only have a few contracts. The performance is not a factor here. As I said, sort is not wrong, but I think `orderBy(filteredContracts, 'startDate')` is more readable.

PR #63, opened 2021-03-16
198 Ability to lock user in HSL ID

Divinare:
Closes https://github.com/HSLdevcom/bultti/issues/198

* Steps for thesting can be found from the issue description.

* when user is logged and then blocked from HSL ID, he can still do things like saving (until he presses f5)
   * there is nothing we can do about this (or we could fetch groups from HSL ID with every request but that's not what we want to do)
   * BUT if you make user.role = 'BLOCKED' by hand in DB, then saving should be prevented -> test that this works (works for me at least)
 
* when user is blocked, it gives 400 (Bad Request) and that seems to crash the app. Maybe not in scope of this task to fix it but maybe it should be fixed somewhere (or here)?

* Note: from now on, user must be 'blocked' in HSL ID (if making it only 'blocked' in DB, it will be overridden when the user logins if he is not in 'blocked group')

### DIFF
--- a/src/util/useAuth.ts
+++ b/src/util/useAuth.ts
@@ -121,5 +121,5 @@ export const useAuth = (): [AuthState, boolean] => {
     }
   }, [currentUser, codeUrlParam, authState, login, refetchUser, navigateNext])
 
-  return [authState, isLoggingLoading]
+  return [authState, isLoginLoading]
 }
--- a/src/util/useAuth.ts
+++ b/src/util/useAuth.ts
@@ -23,7 +23,7 @@ export const useAuth = (): [AuthState, boolean] => {
   // To prevent unwanted navigation, only set this to true when the app should
   // navigate away from the login screen.
   let shouldNavigate = useRef(false)
-  const [login, { loading: isLoggingLoading }] = useMutationData<User>(loginMutation)
+  const [login, { loading: isLoginLoading }] = useMutationData<User>(loginMutation)
   const {
     data: fetchedUser,
     refetch: refetchUserCb,
--- a/src/util/useAuth.ts
+++ b/src/util/useAuth.ts
@@ -23,8 +23,7 @@ export const useAuth = (): [AuthState, boolean] => {
   // To prevent unwanted navigation, only set this to true when the app should
   // navigate away from the login screen.
   let shouldNavigate = useRef(false)
-  const [login, { loading: isLogingLoading }] = useMutationData<User>(loginMutation)
-  const [isLoginFailed, setIsLoginFailed] = useState<boolean>(false)
+  const [login, { loading: isLoggingLoading }] = useMutationData<User>(loginMutation)
   const {
     data: fetchedUser,
     refetch: refetchUserCb,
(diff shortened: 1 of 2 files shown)


src/util/useAuth.ts
```diff
@@ -18,17 +19,17 @@ export enum AuthState {
 export const useAuth = (): [AuthState, boolean] => {
   const [authState, setAuthState] = useState<AuthState>(AuthState.UNAUTHENTICATED)
   const [currentUser, setCurrentUser] = useStateValue('user')
+  let [, setErrorMessage] = useStateValue('errorMessage')
   // To prevent unwanted navigation, only set this to true when the app should
   // navigate away from the login screen.
   let shouldNavigate = useRef(false)
-
-  const [login, { loading: loginLoading }] = useMutationData<User>(loginMutation)
+  const [login, { loading: isLogingLoading }] = useMutationData<User>(loginMutation)
```
danieldunderfelt, line 26: Typo?

src/util/useAuth.ts
```diff
@@ -115,10 +120,11 @@ export const useAuth = (): [AuthState, boolean] => {
           refetchUser()
         } else {
           setAuthState(AuthState.UNAUTHENTICATED)
+          setIsLoginFailed(true)
```
danieldunderfelt, line 123:
```suggestion
          setErrorMessage(text('login_failed'))
```
Can't the message be set here? A new state variable for this seems excessive.

danieldunderfelt (changes requested):
A few comments, otherwise good!

src/util/useAuth.ts
Divinare: Hmm true! This code changed a bit on the way and didn't think setting it here after setIsLoginFailed variable was made!

Divinare:
Review fixes done

src/util/useAuth.ts
```diff
@@ -18,17 +19,16 @@ export enum AuthState {
 export const useAuth = (): [AuthState, boolean] => {
   const [authState, setAuthState] = useState<AuthState>(AuthState.UNAUTHENTICATED)
   const [currentUser, setCurrentUser] = useStateValue('user')
+  let [, setErrorMessage] = useStateValue('errorMessage')
   // To prevent unwanted navigation, only set this to true when the app should
   // navigate away from the login screen.
   let shouldNavigate = useRef(false)
-
-  const [login, { loading: loginLoading }] = useMutationData<User>(loginMutation)
+  const [login, { loading: isLoggingLoading }] = useMutationData<User>(loginMutation)
```
danieldunderfelt, line 26:
```suggestion
  const [login, { loading: isLoginLoading }] = useMutationData<User>(loginMutation)
```

src/util/useAuth.ts
```diff
@@ -115,10 +115,11 @@ export const useAuth = (): [AuthState, boolean] => {
...
         } else {
           setAuthState(AuthState.UNAUTHENTICATED)
+          setErrorMessage(text('login_failed'))
         }
       })
     }
   }, [currentUser, codeUrlParam, authState, login, refetchUser, navigateNext])
 
-  return [authState, loginLoading]
+  return [authState, isLoggingLoading]
```
danieldunderfelt, line 124:
```suggestion
  return [authState, isLoginLoading]
```

danieldunderfelt (approved):
LGTM! The typo fix went in a bit of a wrong direction, but I'll just merge the suggested changes.

PR #64, opened 2021-03-19
WIP: 235 contract validation fixes

Divinare:
Closes https://github.com/HSLdevcom/bultti/issues/235

* So the idea here is to "lock" contracts being used, before we used startDate - endDate which is why contract in a PreInspection !== DRAFT could be swapped on the fly

* Btw, there is some cache bug when making inspection -> in_review -> in_draft, it says:
  "Joiltakin kohteilta puuttuu suoritevaatimus."
  "Joiltakin kohteilta puuttuu suoritevaatimusten kalusto-osuuksia."
---> Press F5 and after that, you should be able to publish again (after the loader spin ends :P)

<-- global mode -->
When attaching procurementUnit to a contract
 - update: contract.procurementUnits
 - update: procurementUnit.contracts

<-- attach contract to inspection mode -->
When saving inspection (draft -> in_review) with procurementUnits
 - update: inspection.procurementUnitContractsMap
 - update: inspection.procurementUnit

<-- attach contract to inspection mode -->
When saving inspection (in_review -> draft)
 - update: inspection.procurementUnitContractsMap = {}

* Had to put error message to one place. With the current situation, frontend doensn't know what error message is thrown. If we could send an error message key from backend, we could use that key to display the error message in the correct place. Made this card priority 1 due to this: https://github.com/HSLdevcom/bultti/issues/227 (see my note there starting with Contract edit)
   * Other option would be to make 2 save buttons & two separate resolvers, one for "sopimusehdot", one for contract form. Imo this could be better with the current code implementation (ItemForm doesn't really support displaying error messages where you want and yes, you could make it work but it makes ItemForm more and more complex...)


danieldunderfelt:
Testing notes:

- UI could also validate that the startDate < endDate and vice versa

src/contract/ContractEditor.tsx
```diff
@@ -149,7 +152,11 @@ const renderInput = ({
     )
   }
 
-  if (key === 'procurementUnitIds' && contract.id && contract.startDate && contract.endDate) {
+  if (key === 'procurementUnitIds') {
+    if (isNew) {
+      return <div />
```
danieldunderfelt, line 157:
```suggestion
      return <React.Fragment />
```

Is the way to return an empty thing which is still a React element.

src/contract/ContractProcurementUnitsEditor.tsx
```diff
@@ -123,7 +115,10 @@ const ContractProcurementUnitsEditor = observer(
         let allSelectedArr = unitOptions.map((unit) => unit.id)
         onChange(allSelectedArr)
       } else {
-        onChange([])
+        let unselectableUnitOptionIds = unitOptions
+          .filter((unit) => unit.isUnselectingDisabled)
+          .map((unit) => unit.id)
```
danieldunderfelt, line 120: I was able to deselect units while they were included in an inspection which was IN_REVIEW. I suggest moving this part to the start of the component and checking `unselectableUnitOptionIds` always when toggling a unit. Now there is no check in the toggle handler if the unit can be toggled.

danieldunderfelt (changes requested):
I had an inspection that was IN_REVIEW, but I could still change the dates of the contract. This rendered the inspection invalid.

The UI should do the following:
- block changing of dates that would mess with any non-draft inspections
- UI validations for startDate < endDate (you can use the min/max props of the date input)

I also got the dirty-form navigation alert after saving units on a new contract.

src/contract/ContractProcurementUnitsEditor.tsx
danieldunderfelt: Although I'm pretty sure I created the inspection while being on the PR branch, it didn't have the unit-contract map. I created a new inspection and then it worked as it should have!

I would still suggest the change, but it's not very critical.

Divinare:
New features need to be implement in this task:
3) Contract dates cannot overlap
4) Make contract startDate as uneditable

Divinare:
The spec for implenting things related to this is the air, it seems not smart trying to iterate all of this when we aren't even sure of how things should go. This is why we should stop here (leave it in a stable state though!)

--> Left for future tasks:
1) block changing of dates with any non-draft inspections
  - maybe make startDate as uneditable
  - get minimum endDate for contract
2) tests to "get contract date candidates"

What was done here?
0) Review fixes done
1) Contract's cant overlap by clicking "Sopimuksessa mukana"
 - HOWEVER, if you edit contract dates, this can cause contracts to overlap. I think we need lock specbefore going any further with this.
2) You have a temprorary button to init inspection.procurementUnitContractMap

PR #65, opened 2021-03-25
205 linked inspection

danieldunderfelt:
Issue: https://github.com/HSLdevcom/bultti/issues/205

Server PR: https://github.com/HSLdevcom/bultti/pull/238

Improvements to linked inspections.

- Show all linked inspections in UI
- Show in UI when linked inspections are updated
- Show confirm prompt when updating linked inspections
- Refetch observed execution requirement when updating linked inspections
- Show validation errors concerning linked inspections


src/postInspection/LinkedInspections.tsx
```diff
@@ -0,0 +1,117 @@
...
+        {
+          query: observedExecutionRequirementsQuery,
+          variables: { postInspectionId: inspection?.id || '' },
+        },
+      ],
+    }
+  )
+
+  let onUpdateConnectedInspection = useCallback(() => {
+    if (isEditable && confirm(text('inspection_editor_linkedInspectionsUpdateConfirm'))) {
```
Divinare, line 72: No need for isEditable check here.

Also do we have a convention of using "on" before a function name if it's getting triggered by onClick? I'm not sure if I personally like that. I feel like it makes it tempting to declare small functions that just call other function when you could call them straight at onClick={}

src/postInspection/LinkedInspections.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+  let onClickConnectedInspection = useCallback(
+    (inspectionId) => {
+      goToPreInspectionReports(inspectionId, InspectionType.Pre)
+    },
+    [goToPreInspectionReports]
+  )
+
+  let [updateConnectedInspections, { loading: updateLoading }] = useMutationData(
+    updateLinkedInspectionsMutation,
```
Divinare, line 56: Hmm, was thinking that could you just add:
```
      inspectionMappings {
        id
        startOfWeek
        inspection {
          ...LinkedInspectionFragment
        }
      }
```
into updateLinkedInspectionsMutation? Then you wouldn't have to have `inspectionQuery` in refetchQueries, am I right?

Divinare (changes requested):
FE review done!

Wasn't able to test it completely since HFP-loading isn't currently really feasible locally. I would hope to make 2x 1 week long preInspections and then make a postInspection using those two. But this is something that is left for us to test in the future :/

Other thing that wasn't tested because currently data is too large in Helsingin Bussiliikenne (we really need that smaller test data!):
If you replace the current in production preInspection, do you get the newer preInspection linked with update linked inspection button? This is what I couldn't test either.

Testing:

- Change the notification into "Oletko varma, että haluat päivittää linkitetyt tarkastukset? Tämän jälkitarkastuksen suoritevaatimusten mahdolliset "% Osuus" -sarakkeeseen tehdyt muutokset resetoidaan.

Bugs found not related to this card:
- Noticed that you can only edit "% Osuus". However, when tried double clicking other cells, "Tallenna muutokset" bar appeared. This shouldn't appear if changes arent dirty.
- Also noticed that if I click uneditable cell and then click "Tallenna muutokset" the app crashes with error:
Variable "$updateValues" got invalid value { id: "76427b9c-9eb3-4419-afab-f8b68c58c525", emissionClass: 7, sanctionThreshold: 0 } at "updateValues[0]"; Field "sanctionThreshold" is not defined by type "ObservedRequirementValueInput"
  - Also this bug happened sometimes (not always) when I edited "% Osuus" and saved

src/postInspection/LinkedInspections.tsx
```diff
@@ -0,0 +1,117 @@
...
+              size={ButtonSize.SMALL}>
+              <Text>update</Text>
+            </Button>
+          )}
+        </div>
+      </Heading>
+      {connectedPreInspections.map((inspectionMapping) => (
+        <InspectionIndexItem
+          key={inspectionMapping.id}
+          onClick={() => onClickConnectedInspection(inspectionMapping.inspection.id)}
```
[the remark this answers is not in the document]
danieldunderfelt: `useInspectionReports` is an abstraction for the navigation function that opens the report page for the inspection. You can name the return value anything you want, and I think `goToPreInspectionReports` makes it crystal clear what will happen when it is clicked.

We can change it to `navigate`, I do like that better, but `navigateToInspection` is not really correct as such because it specifically opens the REPORT page for the inspection. I'll change it to `navigateToInspectionReports`.

And yes, `onClickConnectedInspection` *does* describe what it does! You immediately see that it will be called from a click event, before even seeing the JSX. Then you read what's inside and you see that it navigates to the inspection reports. I can't imagine a clearer function.

And yes I want to keep it as a separate function. It does a distinct _thing_, and it doesn't matter that the thing can be done in one line of code. Inline functions should usually be refactored into separate functions (I haven't of course done it 100% of the time, but this is the better practise IMO). What if that needs to be called from two different places? Then you have two separate inline functions that do the same thing.

src/postInspection/LinkedInspections.tsx
danieldunderfelt: The isEditable check doesn't hurt either! I like to cover as many angles as possible.

And yes, I find it's a nice convention. As I said before, you see immediately that it will be called from a click event, and I prefer functions be extracted into separate functions. Inline functions are mostly only good for prototyping (or laziness) and not the best practise.

src/postInspection/LinkedInspections.tsx
danieldunderfelt: I don't think Apollo is smart enough. `refecthQueries` is not a problem!

danieldunderfelt:
Agreed fixes done!

src/postInspection/LinkedInspections.tsx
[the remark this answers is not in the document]
Divinare: If i say "onClickButtonX()" it doesn't say what it does. Does it open a page? Does it increment a value? No one knows.

I see 1 line function calls fine even if it was in multiple places. At least I instantly see what is being called. I like having a functions that do something rather than just call another function without knowing anything what the function does without looking into it.

Divinare:
Good, navigateToInspectionReports is a lot clearer! Answered to some comments.

Also how about the bugs? I think we might want to make a separate card to fix them

src/postInspection/LinkedInspections.tsx
danieldunderfelt: Again, `refetchQueries` is not a problem and I don't trust Apollo to handle this properly, even if it claims to do so.

src/postInspection/LinkedInspections.tsx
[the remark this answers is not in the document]
danieldunderfelt: It described it's _purpose_, not necessarily exactly what it does. Generally it should only call other functions and not do too much itself, like a controller. I admit I am not always good with refactoring these and they may sometimes be a bit large, but that's the idea.

danieldunderfelt:
I'll have a look at the bugs next week!

danieldunderfelt:
Now fixed/updated:

- Confirm text and update button texts
- Update button style (and remove separate message)
- Quick fixes for requirement table edit
- GraphQL schema updates

> Also found one more minor bug; when I create a new postInspection, "Päivitä linkitetyt tarkastukset" text is red -> shouldnt it show as "up-to-date" as when I just created a new postInspection.

This happened because the stack identifier of the pre-inspection got undefined values, something I fixed previously in this PR. Check the inspections table and see if your pre-inspections have stack identifiers like "12_SY2020_undefined_undefined" and remove the undefined parts. Then it should work! I confirmed that pre-inspections created when on this branch does not have this problem.

src/postInspection/LinkedInspections.tsx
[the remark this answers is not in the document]
Divinare: I think this comes down to a preference. I like less jumping between lines when reading code. Like in ItemForm, I don't like having a separate render functions for label & input rendering.

But we can go which way you want here. This wasn't a big issue here, just more like "a generic code style note"

src/postInspection/LinkedInspections.tsx
Divinare: Yeah we can go with this. When a func does do more than just calling another func, it's OK for me to make it as a separate function. Not sure if it's a good practise, it all comes down to readability!

src/postInspection/LinkedInspections.tsx
Divinare: Okay we shall not trust Apollo here then :D

Divinare:
Seems like everything it's working now!

Review is ok here. Only what I would suggest is adding a prompt to changing inspection startDate & endDate "Oletko varma, että haluat vaihtaa tarkastusjaksoa? Linkitetyt tarkastukset päivitetään ja suoritevaatimukset lasketaan uudestaan linkattujen ennakkotarkastusten pohjalta"

Divinare (approved):
Approving, only one additional suggestion (can be merged after implementing that...)

PR #67, opened 2021-04-01
215 fix execution requirement total rounding

Divinare:
Closes https://github.com/HSLdevcom/bultti/issues/215

Created a report/constants file. I think this is a great report specific config file where that we can use to config all different decimals configs for example.


danieldunderfelt (approved):
Nice! Works well. LGTM

PR #69, opened 2021-04-06
220 sanctions data display

danieldunderfelt:
Closes: https://github.com/HSLdevcom/bultti/issues/220

Server PR: https://github.com/HSLdevcom/bultti/pull/246

- Disable sanction reason grouping
- Show value that triggered sanction in table
- Feature to resize table columns. To use, press down on a table header column and drag sideways. Works on all tables in the UI unless specifically disabled.
- Refactor table component into separate hooks
- Fixes and updates along the way


src/common/table/useColumnResize.ts
```diff
@@ -0,0 +1,157 @@
...
+
+          // Only positive values allowed
+          if (pxWidth >= minWidth && pxWidth <= maxWidth) {
+            nextWidths.splice(colIdx, 1, nextColumnWidth)
+          }
+
+          colIdx++
+        }
+
+        setColumnWidths(nextWidths)
```
Divinare, line 116: Didn't go through this code, as long as it works and it's encapsulated we are happy :+1:

src/common/table/Table.tsx
```diff
@@ -123,196 +131,42 @@ const Table = observer(
...
+      items,
+      sort,
+      setSort,
+      itemsAreSorted: typeof propSort !== 'undefined',
+    })
 
-        widths.push(Math.ceil(colWidth))
-        colIdx++
-      }
+    let { columnNames, columns, rows } = useTableRows<ItemType, EditValueType>({
```
Divinare, line 156: I like the encapsulation here. Making use of plenty of variables and outputing only a few, good job!

Divinare (changes requested):
The table is now a lot more readable when it's nicely split into files where each file describes it's job. This is what I like to see!

Here a few proposals:
- Rounding is a bit off, seeing: 3.3000000000000003

- checkbox proposal: make it a bit bigger and center it (screenshot)
checkbox_proposal

^^^ Note: I increased the size of the checkbox with: transform: scale(1.5);

- About "Sanktiointiin johtava arvo"; I think we could add a tooltip to it. Especially when seeing just "7" there, it doesn't indicate much. Well, that could be casted as "7" -> "Väri". But I would still add an additional tooltip with hovering on it. For example, if I hovered on "Väri", it could display "Ajoneuvoa ajettiin väärällä värillä."

Instead if "R" I would say "V" and when hovered smth like "Tarkasteltu kaluston tyyppi oli eri kuin suunniteltu. Tämä"

So just when sanctionableValue is passed to frontend, add a sanctionableTooltip with it? Or it could be a just a SANCTIONALBE_TOOLTIP_KEY that is mapped LATER into fi.json (we have a card for this already I think)

src/util/getTotal.ts
```diff
@@ -13,3 +13,21 @@ export function getTotal<T, Name extends keyof T>(collection: T[], propName: Nam
 
   return reduced.toString()
 }
+
+export function getTotalNumbers(numbers: Array<number | string>) {
```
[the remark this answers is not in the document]
danieldunderfelt: Well, it's a good to have toolbelt function. I see no reason to remove it.

src/common/table/Table.tsx
```diff
@@ -66,8 +71,8 @@ const HeaderCellContent = styled.div`
 `
 
 const TableBodyWrapper = styled.div`
-  width: 100%;
   position: relative;
+  width: 100%;
```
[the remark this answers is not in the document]
danieldunderfelt: Good cal! This we can remove.

src/common/table/Table.tsx
danieldunderfelt: Cheers! This is something I was mindful of indeed. Originally I had `useTableColumns` outside `useTableRows` in <Table />, but putting it into `useTableRows` got rid of two extra variables that it would have returned into <Table />.

danieldunderfelt:
> Here a few proposals:
> Rounding is a bit off, seeing: 3.3000000000000003
> checkbox proposal: make it a bit bigger and center it (screenshot)

Yes, agreed! This issue or a new card?

> About "Sanktiointiin johtava arvo"; I think we could add a tooltip to it. Especially when seeing just "7" there, it doesn't indicate much. Well, that could be casted as "7" -> "Väri". But I would still add an additional tooltip with hovering on it. For example, if I hovered on "Väri", it could display "Ajoneuvoa ajettiin väärällä värillä."

This is a bit of a larger operation. We should indeed translate the sanction reason, and we can also explain the value that led to the sanction, but the user should really know what the value means in conjunction with the sanction reason.

So yes, we can make a card for this, but it goes under the "User guidance" category which isn't that critical, at least for now.

danieldunderfelt:
Fixed the rounding in the table and removed the redundant component. I don't think unused code should be religiously removed, but it is indeed nice to do so!

The tooltip and the translation should be done in a new issue.

Divinare (approved):
Review approved!

I disagree however with having a ton of custom utils. That's because coders forgot about them / don't know about them. Making something as an util only after **you see it being used multiple times** (i'd say like 3+ times) or if you can for sure say "this is going to get used a lot")

I made a card for that hover tooltip idea: https://github.com/HSLdevcom/bultti/issues/248

danieldunderfelt:
> I disagree however with having a ton of custom utils. That's because coders forgot about them / don't know about them. Making something as an util only after you see it being used multiple times (i'd say like 3+ times) or if you can for sure say "this is going to get used a lot"

I disagree with this. If I make a useful utility but end up not needing it, I don't think it's necessary to remove it if I think I might need it in the future. And id'd say it's good to create a general utility immediately if you find yourself writing the same code a second time.

It's a different story with specialised code that ends up redundant. If it's not a generally useful utility it should be removed. Example being the TableBodyWrapper component I accidentally left in.

Divinare:
I agree on those:
* And id'd say it's good to create a general utility immediately if you find yourself writing the same code a second time.
* if it's not a generally useful utility it should be removed"

But it's up to the viewer to see if it's a good utility :D

But yea, this can be merged.

PR #73, opened 2021-04-16
251 inspection view updates

danieldunderfelt:
Closes https://github.com/HSLdevcom/bultti/issues/251

Server PR: https://github.com/HSLdevcom/bultti/pull/253 (nothing but a comment)

- Add inspection name to "timeline" inspection list
- Fix inspection listing performance issues by querying for less fields


src/util/getTotal.ts
```diff
@@ -1,17 +1,19 @@
...
+export function getTotalBig<T, Name extends keyof T>(collection: T[], propName: Name): Big {
+  return collection.reduce((total: Big, item: T) => {
     let propValue = item[propName]
 
-    if (typeof propValue !== 'number') {
+    if (typeof propValue !== 'number' && typeof propValue !== 'string') {
       return total
     }
 
     return total.plus(propValue)
```
Divinare, line 11: Nit: I think it reads slightly better with:
```suggestion
    if (typeof propValue === 'number' || typeof propValue === 'string') {
      return total.plus(propValue)
    }

    return total
```

src/executionRequirement/RequirementEquipmentList.tsx
```diff
@@ -19,24 +21,26 @@ export type PropTypes = {
...
-  model: 'Malli',
-  type: 'Tyyppi',
-  emissionClass: 'Euroluokka',
-  registryDate: 'Rek.päivä',
-  percentageQuota: 'Osuus',
-  kilometerRequirement: 'Kilometriosuus',
-  amount: 'Määrä',
+  model: text('executionRequirement_model'),
+  type: text('executionRequirement_type'),
+  emissionClass: text('executionRequirement_emissionClass'),
```
Divinare, line 38: These are not getting used it seems:
texts

src/equipment/EquipmentList.tsx
```diff
@@ -147,6 +147,12 @@ const EquipmentList: React.FC<PropTypes> = observer(
             return round(getTotal(equipment, 'kilometerRequirement'), 3) + ' km'
           case 'amount':
             return round(getTotal(equipmentGroups, 'amount'), 3) + ' kpl'
+          case 'age':
+            // Show the unweighted average in the total row
+            return (
+              round(getTotalBig(equipmentGroups, 'age').div(equipmentGroups.length || 1), 3) +
```
Divinare, line 153: Should it say e.g. "ka 13.5v"?

ka = keskiarvo

Also hmm, we could have use for weight helper here e.g.
```
'ka ' + round(unweightedAverage(getTotalBig(equipmentGroups, 'age'), equipmentGroups.length), 3) +'v'
```
Then we also don't need the comment "Show the unweighted average in the total row"

Divinare: Also hmm, maybe unweightedAverage could use round underneath but then I would call it like:
```
'ka ' + unweightedAverage(sum: getTotalBig(equipmentGroups, 'age'), count: equipmentGroups.length, decimals: 3) +'v'
```

Divinare (changes requested):
Found some small things that could be improved

src/util/getTotal.ts
danieldunderfelt: Yeah we can do it this way!

src/executionRequirement/RequirementEquipmentList.tsx
danieldunderfelt: Fixed in https://github.com/HSLdevcom/bultti-ui/pull/72

src/equipment/EquipmentList.tsx
danieldunderfelt: Sure, we could do that, but it would only take one argument and get the decimals from constants. I'll do it!

But I still think the comment is needed, since weighted averages is the _norm_ for Bultti, and an average function would not be called "unweighted", as an unweighted average is just an average. I want to make it clear that this number is different from the weighted average.

danieldunderfelt:
> Btw PR could mention that dev environment was already updated with these changes

Dev is for testing and may change frequently. Slack is a much better way to communicate this.

src/util/getTotal.ts
danieldunderfelt: Actually we can't since Typescript doesn't like it. We'd have to check `propValue` much more.

Divinare (approved):

PR #75, opened 2021-04-19
230 prevent season switch when opening inspection

Divinare:
Closes https://github.com/HSLdevcom/bultti/issues/230

As a small bonus, added inspection date range (we've had discussion that it's ok to do):

Tried to override the style at: GlobalSeasonFilter.tsx
with:
const StyledDropdown = styled(SidebarStyledSelect)`
  padding-bottom: 0;
`

but got this error:
ReferenceError: Cannot access 'SidebarStyledSelect' before initialization
so that's why I made a separate styled element in AppSidebar

### DIFF
--- a/src/common/components/AppSidebar.tsx
+++ b/src/common/components/AppSidebar.tsx
@@ -117,7 +117,7 @@ export const SidebarStyledSelect = styled(Dropdown)`
   }
 
   > div {
-    padding: 0 1rem 1rem;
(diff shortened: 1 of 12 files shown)


danieldunderfelt (approved):
Looks good!

I made some additions:

- Added back season check in inspection editor
- Use inspection type strings util to navigate to the correct inspection type page
- Separate module for sidebar dropdown to prevent circular dependencies

PR #78, opened 2021-04-26
216 use thousand separator in tables

Divinare:
Closes https://github.com/HSLdevcom/bultti/issues/216

I'd prefer to just make one util: getThousandSeparatedNumber() to handle all 3 input types: input: number | string | Big

but to implement that, you have to check if input type is Big, but how to do that? This didn't work: typeof input === 'Big'


danieldunderfelt (approved):
I unified the functions as requested. LGTM!

PR #79, opened 2021-04-27
230 notify of season switch when opening inspection

danieldunderfelt:
Update #75 to show a notification when the season changes.

Issue: https://github.com/HSLdevcom/bultti/issues/230

Changes:

- Create new InfoMessages component that can show info and error messages.
- Show info message when the season changes automatically
- Show messages as toasts
- Refactor state and related utilities for more type safety and compatibility


src/util/useShowNotification.ts
```diff
@@ -0,0 +1,16 @@
+import { useStateValue } from '../state/useAppState'
+import { useCallback } from 'react'
+import { UINotification } from '../type/state'
+
+export function useShowNotification(type: UINotification['type']) {
+  let [, { add }] = useStateValue('notifications')
+  return useCallback((message: string) => add({ message, type }), [add])
+}
```
Divinare, line 8: Now we have diff naming:
var setInfoMessage = useShowInfoNotification()
let showErrorNotification = useShowErrorNotification()
var setErrorMessage = useShowErrorNotification()
let showErrorMessage = useShowErrorNotification()

These should be renamed as:
var showInfoNotification = useShowInfoNotification()
let showErrorNotification = useShowErrorNotification()
var showErrorNotification = useShowErrorNotification()
let showErrorNotification = useShowErrorNotification()

But idk, maybe i'd prefer this style:
```
type MessageType = 'info' | 'error'
export const showNotification = ({ message, type = 'info' }: { message: string, type?: MessageType }) => {...}
showNotification({ message: '...', type: 'error'})
```
At least this with approach you don't end up using different names for the function :P which again slightly improves readability. But yea I also agree one can see this as "I prefer this car" type of things

src/text/fi.json
```diff
@@ -169,6 +172,7 @@
   "inspectionPage_sanctions": "Sanktiointi",
   "inspectionPage_relatedUsers": "Käyttäjät",
 
+  "inspection_seasonChangedAutomatically": "The season was automatically changed to ${newSeason}",
```
Divinare, line 175: Finnish?

Divinare (changes requested):
Testing:

Styling looks greatly better now!

I noticed that if you have lets say an existing notification "The season was automatically changed to SY2020", it wont show up again if it's switched back. I would show duplicate messages OR you could flash the existing notification to indicate that it occured again (the latter is what I did in a previous project where it worked well). Maybe here i'd go with the first option however because errors / messages are shown quite infrequently.

Otherwise this seems to work well, a nice addition were those globalState fixes + that there is no longer an empty option in global selectors

src/text/fi.json
danieldunderfelt: Hahah yes... :D

src/util/useShowNotification.ts
danieldunderfelt: Good point about naming, I'll unify the names. When refactoring I just aliased `add` to the name already used in the code to show an error message, which was mostly `setErrorMessage`. This resulted in less errors and less changes, but maybe it's better to use the same name everywhere.

For the hook itself, I like using different functions for showing different types, but the general `useShowNotification` can also be used.

danieldunderfelt:
> I noticed that if you have lets say an existing notification "The season was automatically changed to SY2020", it wont show up again if it's switched back. I would show duplicate messages OR you could flash the existing notification to indicate that it occured again (the latter is what I did in a previous project where it worked well but there the same error happened quite often). Maybe here i'd go with the first option however because errors / messages are shown quite infrequently.

Yes, we should add some nice animations and your suggestions are good, but I wanted to keep it simple for now as this task got way too big already. We can continue working on the notifications in another issue.

danieldunderfelt:
> I noticed that if you have lets say an existing notification "The season was automatically changed to SY2020", it wont show up again if it's switched back. I would show duplicate messages OR you could flash the existing notification to indicate that it occured again (the latter is what I did in a previous project where it worked well but there the same error happened quite often). Maybe here i'd go with the first option however because errors / messages are shown quite infrequently.

Yes, we should add some nice animations and your suggestions are good, but I wanted to keep it simple for now as this task got way too big already. We can continue working on the notifications in another issue.

Divinare:
Created a card for handling duplicated notifications: https://github.com/HSLdevcom/bultti/issues/269

PR #89, opened 2021-05-07
Fixed Inspection fetch after removal + cache bug

Divinare:
Closes https://github.com/HSLdevcom/bultti/issues/270

Two bug fixes here:
* Cache: after removal, fetchPolicy: 'no-cache' was needed because the query used cache and didnt go to your backend.
* After removal, some onRefresh function was being called and this caused EditInspection page to render when it tried fetching the removed pre/post inspection


src/page/SelectInspectionPage.tsx
```diff
@@ -25,6 +25,7 @@ const SelectInspectionPage: React.FC<PropTypes> = observer(({ inspectionType })
     {
       skip: !operatorIsValid(operator) || !seasonIsValid(season),
       notifyOnNetworkStatusChange: true,
+      fetchPolicy: 'no-cache',
```
danieldunderfelt, line 28: Try `cache-and-network` instead. Never use `no-cache`.

danieldunderfelt (changes requested):
The best course of action would probably be to include the `currentInspectionsByOperatorAndSeasonQuery` in the `refetchQueries` of the inspection removal mutation.

Then there is probably no need for a special `fetchPolicy`.

danieldunderfelt:
Added the selectinspection page query as a refetch query to remove, which negates the need for a cache-policy.

PR #91, opened 2021-05-11
271 fix sanction report percentages

Divinare:
Closes https://github.com/HSLdevcom/bultti/issues/271

**Note**: If you want more speed for your testing, make sure to comment this line out from backend's createSanctionSummaryReportData():
```
// let unitExecutionItems = await calculateObservedUnitExecution(inspection)
```

**What was done in this pr:**

* Fixed report queries gql() taken off
* Fixed report cell rendering
* The required minimum to get percentages work correctly.
  * **Total sum of percentages is not working** yet and it couldn't be fixed here (look proposals)
* Renamed "Painotettu keski-ikä" -> "Toteutunut painotettu keski-ikä", it's now "Toteutunut painotettu keski-ikä" in every place.
* Removed total column
* Highlighting cells in "totueutunut painotettu keski-ikä"


danieldunderfelt:
> I would also get rid of sanctionSummaryTransform() in frontend and move the logic in backend. We would return rows that don't have duplicated procurementUnitIds.

> Btw, I know there was this comment "The reason for this is that the sanctionReasons and sanction amounts are
dynamic, and theGraphQL response fields must be regular." - is that the reason why we need sanctionSummaryTransform() at frontend side?

Indeed, not possible to have dynamic columns with GraphQL. There's not really any logic in the UI, it just combines the rows by unit and creates the sanction reason/amount columns.

The point of this report is to mimic an important report that exists as an Excel file in old Bultti.

danieldunderfelt:
> What about making a separate report for each sanction type? It could show each sanction row and the TOTAL numbers below (there is no other way than this to actually show the OperatingAreaName except showing sanction rows invidually in UI)

Current reports already show the things that lead to sanctions, but if the users want a more sanction-focused report for each issue we can certainly do that.

danieldunderfelt:
> Also if BE's createSanctionSummaryReportData() output data was nicer, we could utilize the same method when sending sanctions to KOLA in #59

This function is for creating the report. If a similar function is needed, common logic should be extracted. I don't think this should be used for anything else than creating the report.

src/report/cellHighlight.ts
```diff
@@ -0,0 +1,10 @@
+export const cellHighlight = {
+  sanctionSummary: shouldHighlightCell,
+}
+
+function shouldHighlightCell(rowItem, key: string): boolean {
```
Divinare, line 5: rowItem currently has `any` type but I didn't know how to provide one

src/common/table/TableRow.tsx
```diff
@@ -84,6 +86,13 @@ export const TableRow = observer(
             colIndex={colIndex}
             tabIndex={index * allRows.length + colIndex + 1}
             cell={[key as keyof ItemType, val]}
+            rowHighlightColor={
+              cellHighlighter
+                ? cellHighlighter(rowItem, key as string)
+                  ? '--yellow'
+                  : undefined
+                : undefined
```
Divinare, line 94: I broke row highlighting when a vehicle is added to preInspection unit's equipmentList. Not quite sure how to support that and cellHighlighter at the same time. Didn't have time to look into this yet

src/common/table/tableUtils.tsx
```diff
@@ -27,17 +27,12 @@ export interface IFilteredSortedResponse<DataType extends TableItemType> {
...
-    if (typeof val === 'boolean' || typeof val === 'undefined' || val === null) {
-      return (
-        <span
-          style={{
-            fontSize: '1.2rem',
-            marginTop: '-3px',
-            display: 'block',
-          }}>
-          {val ? '✓' : '⨯'}
-        </span>
```
danieldunderfelt, line 39: Ok, if it works without the styling...

src/common/table/Table.tsx
```diff
@@ -106,6 +107,7 @@ export type TablePropTypes<ItemType extends TableItemType> = {
   ) => React.ReactNode
   getColumnTotal?: (key: string) => React.ReactChild
   highlightRow?: (item: ItemType) => boolean | string
+  cellHighlighter?: (item: TableRowWithDataAndFunctions<ItemType>, key: string) => boolean
```
danieldunderfelt, line 110: I think the cell highlight function should follow the same format as the pre-existing `highlightRow`. It would be more consistent if it were named `highlightCell`.

src/report/ReportContainer.tsx
```diff
@@ -95,6 +96,11 @@ const ReportContainer = observer((props: PropTypes) => {
         continue
       }
 
+      // We don't want to display unitEquipmentMaxAge column in sanction report
+      if (colName === 'unitEquipmentMaxAge') {
+        continue
+      }
```
danieldunderfelt, line 102: Could be combined with the continue condition above.

src/report/cellHighlight.ts
danieldunderfelt: This function is very specific to the sanction summary report, so it should be named accordingly (eg. `shouldHighlightSanctionSummaryReportCell`). Then you can also set a static type: `TableRowWithDataAndFunctions<SanctionSummaryReportData>`.

src/report/cellHighlight.ts
```diff
@@ -0,0 +1,10 @@
+export const cellHighlight = {
```
danieldunderfelt, line 1: Maybe `reportCellHighlightMap` would be a more descriptive name?

src/common/table/Table.tsx
danieldunderfelt: It could also return a string like `highlightRow` to enable highlighting cells with different colors. But it's an idea for the future.

danieldunderfelt (changes requested):
Good work, but left some minor comments!

src/report/ReportContainer.tsx
Divinare: For better readability I think separate if is better... Now you can see that the comment relates to this statement

src/report/cellHighlight.ts
Divinare: Yes. Also renamed reportTransforms -> reportTransformsMap

src/report/ReportContainer.tsx
danieldunderfelt: Fair point, either way is fine. Comments can be added to combined iffs too, though.

PR #97, opened 2021-06-07
318 execution requirement values

danieldunderfelt:
Closes : https://github.com/HSLdevcom/bultti/issues/318

Mostly schema updates for HSLdevcom/bultti#318.

Server PR: https://github.com/HSLdevcom/bultti/pull/323

Noticed that table inputs weren't focused when activated and tried to fix this, but couldn't solve it quickly enough. Left mechanism for setting a ref on the table input as that may be useful later.


Divinare (approved):
Works now as expected!

PR #103, opened 2021-06-18
333 role fixes according to role matrix

Divinare:
Closes https://github.com/HSLdevcom/bultti/issues/333

Note: currently only admins can accept pre and post inspections. We'll implement double accept later where also operator can accept their owns inspections

Questions:
* Should I remove those methdos marked with TODO: remove? Note: I don't want this PR to be merged with any new TODO's so something needs to be done with them (I found like 5-8 unused resolvers)
* Is guardPlannedUnitExecutionRequirementResolvers OK that it has allowedRoles: [UserRole.HSL]?
* refreshEquipmentInCatalogueMutation: this was OK to be available for all roles
* Shouldn't we remove DEV Load sanctions button from post inspection's page? We have devResolver for that, if ever want to test running some functions in backend.
* "Sopimuksessa mukana toggle" - is the endpoint of this now correctly restricted by role checks? I guess admins and operators can do it or am I right? Can't easilly find which resolver this checkbox calls....

How to test this PR:

* Test every action/resolver with every role (admin, hsl user, operator). Ensure that each action has access rights according to our access right table: https://gitlab.hsl.fi/bultti/bultti-testaus/bultti/-/wikis/Ka%CC%88ytto%CC%88oikeudet
* You should also test opening post inspections that are in DRAFT with operator and HSL user - shouldn't work.
* Also you shouldn't be able to open pre inspections in DRAFT with HSL user

### DIFF
--- a/src/inspection/SelectInspection.tsx
+++ b/src/inspection/SelectInspection.tsx
@@ -1,7 +1,7 @@
 import React, { useCallback } from 'react'
 import styled from 'styled-components/macro'
 import { observer } from 'mobx-react-lite'
-import { Inspection, InspectionType } from '../schema-types'
+import { Inspection, InspectionType, UserRole } from '../schema-types'
 import { orderBy } from 'lodash'
 import { FlexRow } from '../common/components/common'
 import { useStateValue } from '../state/useAppState'
@@ -17,7 +17,7 @@ import InspectionCard from './InspectionCard'
 import { Text } from '../util/translate'
 import { operatorIsValid } from '../common/input/SelectOperator'
 import { seasonIsValid } from '../common/input/SelectSeason'
-import { useHasAdminAccessRights, useHasOperatorUserAccessRights } from '../util/userRoles'
+import { useHasAccessRights } from '../util/userRoles'
 
 const SelectInspectionView = styled.div`
   position: relative;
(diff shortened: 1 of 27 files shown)


src/inspection/SelectInspection.tsx
```diff
@@ -93,13 +94,24 @@ export type PropTypes = {
...
+    let allowedRoles: UserRole[] = []
+    if (inspectionType === InspectionType.Pre) {
+      allowedRoles = [UserRole.Admin, UserRole.Operator]
+    } else {
+      allowedRoles = [UserRole.Admin]
+    }
+    let canCreateInspection = useHasAccessRights({
+      allowedRoles,
+      operatorId: globalOperator?.id,
+    })
```
danieldunderfelt, line 109: Same code duplicated a few times. Could probably be extracted into a separate function.

src/inspection/inspectionUtils.ts
```diff
@@ -236,3 +238,60 @@ export function isPreInspection(inspection: unknown): inspection is PreInspectio
...
+  } else {
+    allowedRoles = [UserRole.Admin, UserRole.Hsl, UserRole.Operator]
+  }
+
+  return hasAccessRights({
+    user,
+    allowedRoles,
+    operatorId,
+  })
+}
```
danieldunderfelt, line 274: This should be made to work and used in places where the same kind of code is duplicated.

src/inspection/inspectionUtils.ts
```diff
@@ -236,3 +238,60 @@ export function isPreInspection(inspection: unknown): inspection is PreInspectio
...
+  } else {
+    allowedRoles = [UserRole.Admin]
+  }
+
+  return hasAccessRights({
+    user,
+    allowedRoles,
+    operatorId,
+  })
+}
```
danieldunderfelt, line 297: Same for this function.

danieldunderfelt (changes requested):
Good work, left some suggestions.

danieldunderfelt:
- Should I remove those methdos marked with TODO: remove? Note: I don't want this PR to be merged with any new TODO's so something needs to be done with them (I found like 5-8 unused resolvers)
  - We can clean up unused stuff in another PR.
- Is guardPlannedUnitExecutionRequirementResolvers OK that it has allowedRoles: [UserRole.HSL]?
  - I can't find anything called this in either repo.
- Shouldn't we remove DEV Load sanctions button from post inspection's page? We have devResolver for that, if ever want to test running some functions in backend.
  - Sure
- "Sopimuksessa mukana toggle" - is the endpoint of this now correctly restricted by role checks? I guess admins and operators can do it or am I right? Can't easilly find which resolver this checkbox calls....
  - It uses the `modifyContract` mutation. Only allowed for admins currently.

danieldunderfelt (approved):
LGTM

PR #116, opened 2021-07-09
370 sanctions by scope

danieldunderfelt:
Closes HSLdevcom/bultti#370

Group report tables by some value if the groupRowsBy value is set. Currently set for the "all sanctions" report.

Server PR: https://github.com/HSLdevcom/bultti/pull/390


src/common/table/useTableColumns.ts
```diff
@@ -12,11 +13,16 @@ export function useTableColumns<ItemType extends TableItemType>({
   columnOrder?: string[]
   hideKeys?: string[]
 }) {
+  // Collect an array of keys that are empty for all items. These will be hidden.
+  let emptyValues = useMemo(() => findEmptyKeys(items), [items])
```
Divinare, line 17: So if all values of a column are empty, the whole column itself disappears? I dont see the point

src/report/ReportContainer.tsx
```diff
@@ -168,14 +168,22 @@ const ReportContainer = observer((props: PropTypes) => {
...
-        />
+        preparedReport && (
+          <FilteredResponseTable
+            data={preparedReport}
+            tableState={tableState}
+            columnLabels={columnLabels}
+            keyFromItem={reportKeyFromItem}
+            getColumnTotal={calculateReportTotals}
+            getCellHighlightColor={getCellHighlightColor}
+            groupBy={
```
Divinare, line 179: Can you just say:
```
groupByKey={report?.groupRowsBy ? text(item[report.groupRowsBy] as string) : 'default'}
```

using just a string is simplier, function not needed at this point?

src/text/fi.json
```diff
@@ -287,6 +287,10 @@
   "report_filtering_apply": "Suodata",
   "report_filtering_hint": "Suodata raporttitaulukkoa lisäämällä kenttä ylläolevalla painikkeella. Kirjoita tekstikentään arvo ja valitse valikosta sarakkeen jonka mukaan haluat suodata tuloksia. Jos haluat suodata kyllä/ei arvoja, voit käyttää suodatinarvona: true/false, kyllä/ei tai yes/no. Negatiiviset arvot toimii myös jos haluat listata rivit mistä puuttuu jokin arvo.",
 
+  "DEPARTURE": "Lähdöittäin",
+  "OPERATING_AREA": "Alueittain",
+  "PROCUREMENT_UNIT": "Kohteittain",
```
Divinare, line 292: report_group_by_procurement_unit: "Kohteittain"

And when this key is used, say something like
```
text[report_group_by_${object.toLowerCase()}]
```

src/report/ReportContainer.tsx
```diff
@@ -168,14 +168,22 @@ const ReportContainer = observer((props: PropTypes) => {
...
-          data={preparedReport}
-          tableState={tableState}
-          columnLabels={columnLabels}
-          keyFromItem={reportKeyFromItem}
-          getColumnTotal={calculateReportTotals}
-          getCellHighlightColor={getCellHighlightColor}
-        />
+        preparedReport && (
+          <FilteredResponseTable
+            data={preparedReport}
```
Divinare, line 173: I feel like this code could be improved a bit - now there are many variables declared in this scope that are only used by FilteredResponseTable e.g. columnLabels, tableState, reportKeyFromItem, calculateReportTotals, getCellHighlightColor. Could FilteredResponseTable take data in and declare those things including preparedReport inside FilteredResponseTable?

Divinare (changes requested):
Testing:

Issue 1:

- Kilpakohteittain is missing from "Lista kaikista sanktioista" at tulokset tab
  - I guess it's missing because it's empty. But you should still show the empty table to indicate that "Kilpakohteittain" exists.


Issue 2:

Alueittain:
Docs say that you should have columns "seuranta-alue, sanktio%, sanktion määrä (€) ja sanktion syy."

- is seuranta-alue the same as sanktion piiri?

Lähdöittäin:
- is variantti missing or is it column "reitti"? Docs say "Linja & variantti"

Issue 3:

When I download the excel, I did get all of the sanctions "kohteittain" but none of "alueittain" (had 1 row of alueittain but it didn't appear on the downloaded excel"

src/common/table/useTableColumns.ts
danieldunderfelt: The point is to hide departure-related columns for the tables with sanctions that are not departure-scoped. No need to have routeId etc for area and unit sanctions and they would just take up space.

And in general I prefer that all columns have a function, otherwise they are unnecessary.

src/report/ReportContainer.tsx
danieldunderfelt: Simpler yes, but less flexible. I don't see why this is a problem.

src/text/fi.json
danieldunderfelt: I'd like to be able to feed the "raw" text to the translate function and get a translation back. We can prepend `report_group_by_` sure, but I don't really see the point of complicating it further.

src/report/ReportContainer.tsx
danieldunderfelt: FilteredResponseTable is a general component whose task is to render a table based on an object like what "preparedReport" contains. You do have a point though, and we can make an intermediary component that crunches the data for `FilteredResponseTable` which the execution requirement components don't need.

danieldunderfelt:
Issue 1:

This suggestion would needlessly complicate the logic, since you would have to essentially hardcode the categories in order to show empty ones. Now it dynamically groups the sanctions by a key, and the groups themselves can be added to or removed without Bultti caring about exactly what they are. And all of this just to show an empty table, which I don't see the point of doing to begin with.

Issue 2:

Yes, "sanktion piiri" is the one. As I said, don't read these like the bible. There is room to be creative to make a solution that matches the specs and does not require too much work to implement. In this case, the Sanction entity has the `sanctionScope` prop which is used for all three groups, so putting "Seuranta-alue" as the translation here would make no sense for the two other groups.

And yes "reitti" is a combination of the line and the variant.

Issue 3:

There are two pages in the excel, each sanction group has its own page.

src/report/ReportContainer.tsx
Divinare: Not a problem but it's a bit better code. Minor issue though

src/text/fi.json
Divinare: Thing is the scope is there so that you can easilly see where it's being used. Now your best shot is to search "Lähdöittäin" from UI to see where it's being used... If it was just "lähtö" or "kohde" then this would be ok

Divinare:
Yeah my fault, didn't see the second page :D

I see, empty table would be good but the current code doesn't support it very well. Should tell UI to render specific column names without data to show.... We can leave it for now

src/common/table/useTableColumns.ts
Divinare: I see the point.

But what if you have only one row in some report where a column has no data. Will the column name without data disappear?

PR #117, opened 2021-07-12
367 JOLA rows loading state fix

danieldunderfelt:
Closes HSLdevcom/bultti#367

Add better loading state to JOLA rows box in post-inspection. Hide "not found" message until loading completes.


Divinare (changes requested):
The loading works well! Also seems like "suoritevaatimus" loading is faster now!

One thing I noticed, if you switch inspectionDate, JOLA doesn't reload - user has to manually click "päivitä"

Shouldn't it reload JOLA sanctions if inspectionDate is changed (when saving a new inspectionDate)?

danieldunderfelt:
> Shouldn't it reload JOLA sanctions if inspectionDate is changed (when saving a new inspectionDate)?

Good idea, I'll add this!

Divinare:
Good, changes seems to work

PR #122, opened 2021-07-27
#410 route deviation report schema update

danieldunderfelt:
Schema updates for server PR: https://github.com/HSLdevcom/bultti/pull/420

Closes HSLdevcom/bultti#410


Divinare:
Testing:

Issue 1:

Some registry numbers are missing from allDeviations excel:
registry_numbers_missing

Divinare:
Btw shouldn't excel show the same table columns as in UI? Currently not seeing "Rekisterinumero" column in UI

danieldunderfelt:
> Btw shouldn't excel show the same table columns as in UI? Currently not seeing "Rekisterinumero" column in UI even though it's in the excel downloaded excel.

Yes, but the rows of deviation pairs which were planned by HSL have no registry number. The excel report does not filter out columns where no rows have values, like the UI does.

Divinare:
For me, in "Linjasivurikkeet (HSL+liikennöitsijä)" report UI shows 32 rows and downloaded excel sheet only 8 rows. There seems to be some kind of bug, the rows in UI and downloaded excel should match right?

danieldunderfelt:
That sounds like a bug! I'll check it out.

danieldunderfelt:
> For me, in "Linjasivurikkeet (HSL+liikennöitsijä)" report UI shows 32 rows and downloaded excel sheet only 8 rows. There seems to be some kind of bug, the rows in UI and downloaded excel should match right?

The Excel file has pages for each day type. All the rows seem to be present.

PR #125, opened 2021-08-04
424 unit settings and max age calculation

danieldunderfelt:
Closes https://github.com/HSLdevcom/bultti/issues/424

Add option period start date and option max age increase method fields to procurement unit edit form. Update the data shown in the procurement unit expanding box header. Calculate the max age of the units using the new props. Create and update translations regarding these features.

Server PR: https://github.com/HSLdevcom/bultti/pull/428


src/procurementUnit/procurementUnitUtils.ts
```diff
@@ -0,0 +1,48 @@
...
+import { differenceInMonths, differenceInYears, isBefore } from 'date-fns'
+import { round } from '../util/round'
+
+// Calculate the maximum age value of the unit. It starts with the base "maximumAverageAge" value, and
+// if the unit uses option time, the maximum age is increased. There are tqo methods for increasing
+// the maximum age value, by 0.5 years per option year or monthly. This function calculates
+// the increase based on when the option period started and the reference date provided.
+// This same function also exists on the server, so if you touch this, also make the
+// changes in the server function of the same name.
+export function calculateMaximumAverageAge(
```
Divinare, line 12: I strongly suggest calculating this value only in BE and using procurementUnit.maximumAverageAge

Divinare (changes requested):
Code review done! Didn't test properly yet

src/text/fi.json
```diff
@@ -108,8 +108,15 @@
...
-  "procurementUnit_ageRequirementCalculated": "Laskettu maksimi keski-ikä vaatimus + optiovuodet",
+  "procurementUnit_ageRequirement": "Maksimi keski-ikä",
+  "procurementUnit_noOptions": "Ei optiota",
+  "procurementUnit_optionsUsed": "Käytetyt optiot",
+  "procurementUnit_optionPeriodStart": "Optio-kauden alku",
+  "procurementUnit_optionMaxAgeIncreaseMethod": "Maksimi keski-iän korotustapa",
+  "procurementUnit_optionMaxAgeIncreaseMethod_NO_INCREASE": "Ei korotusta",
+  "procurementUnit_optionMaxAgeIncreaseMethod_MONTHLY_INCREASE": "Korotetaan kuukausittain",
+  "procurementUnit_optionMaxAgeIncreaseMethod_HALF_YEAR_INCREASE": "Korotetaan 0.5v / optio-vuosi",
+  "procurementUnit_ageRequirementWithOptions": "Maksimi keski-ikä optiokorotuksilla",
```
Divinare, line 119: Suggestion:
Maksimi keski-ikä optiokorotuksilla (tällä hetkellä)

For me it's not quite clear that the current date affects the value. In future, after we pass the option year start date, option max age starts increasing

Divinare:
Testing:

The numbers seems to be incorrect. As testing in 17.8.2021:

Optio-kauden alkupvm 16.8.2019 should be 8,5, was 8
Optio-kauden alkupvm 17.8.2019 should be 8,5, was 8
Optio-kauden alkupvm 18.8.2019 should be 8 , was 8

src/procurementUnit/procurementUnitUtils.ts
Divinare: Seems like this value is really wanted in UI. How about making a resolver and getting this value that way from backend?

Divinare:
Review done!

src/procurementUnit/procurementUnitUtils.ts
danieldunderfelt: Yes, good idea. Fixed!

src/text/fi.json
danieldunderfelt: Might be a bit long... maybe as a ? text.

danieldunderfelt:
> A) The numbers seems to be incorrect. As testing in 17.8.2021

The option period is never compared to the _current_ date. When on the "Muokkaa kilpailukohteet" page, the option period end date is the start of the season (for lack of a better thing to use). When on the edit pre-inspection page, the option period end date is the start of the inspection period.

src/text/fi.json
```diff
@@ -108,8 +108,15 @@
   "procurementUnit_kilometersInWeek": "Kilometrejä viikossa",
   "procurementUnit_unitValidTime": "Kohteen voimassaoloaika",
   "procurementUnit_operationArea": "Seuranta-alue",
-  "procurementUnit_ageRequirement": "Maksimi keski-ikä vaatimus",
-  "procurementUnit_ageRequirementCalculated": "Laskettu maksimi keski-ikä vaatimus + optiovuodet",
+  "procurementUnit_ageRequirement": "Maksimi keski-ikä",
+  "procurementUnit_noOptions": "Ei optiota",
```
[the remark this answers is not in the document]
danieldunderfelt: Sure, I'll use that.

src/text/fi.json
Divinare: Yeah, I guess the current one is OK. It will most likely be changed in the future anyway

Divinare:
The option year should update when calendar is updated. There is an apollo cache issue, otherwise this PR is OK!

Steps something like
1) Set option year date for some procurementUnit that has not been touched
2) Switch season
3) --> option year doesn't up update correctly. It updates once you change option year a bit
4) But now, if switch between seasons, it suddenly works because of apollo caches the calculated option year

---> this should be fixed that it would always calculate and not use apollo cache

Divinare (approved):
LGTM, cache issue seems to be solved!

PR #127, opened 2021-08-09
273 pre-inspection e2e tests

danieldunderfelt:
Closes https://github.com/HSLdevcom/bultti/issues/273

- Fix existing authentication tests
- Add basic tests for creating and publishing pre-inspections
- Add required testId stuff and custom Cypress commands

PR #128, opened 2021-08-10
432 post inspection e2e tests

danieldunderfelt:
Closes HSLdevcom/bultti#432

Basic E2E tests for post-inspections. Based on the pre-inspection tests. Includes fixes and updates. Based on the branch for PR #127 so merge that one first.

Server PR: https://github.com/HSLdevcom/bultti/pull/435


Divinare (changes requested):
A couple of things:

1) I noticed a problem; if I run post_inspection_publish tests, for example "Can revert sanctionable post-inspection to draft", it throws an error:

duplicate key value violates unique constraint "PK_0065efe666665a68166525b637c"


2) The tests seem to create multiple post inspections as i'm running the tests. I would want it to open an existing postInspection if one is available


3) Also notified that currently teasts can't create inspections if "Remove inspections" button is clicked. This should be fixed here: https://github.com/HSLdevcom/bultti/issues/430

danieldunderfelt:
> The tests seem to create multiple post inspections as i'm running the tests. I would want it to open an existing postInspection if one is available

This is how it should work.

Divinare (approved):
Tests seem to work now as contracts and procurementUnits get removed properly! LGTM

PR #134, opened 2021-08-18
User can now be removed from dev page. Also user.role = blocked does not exist anymore

Divinare:
Closes https://github.com/HSLdevcom/bultti/issues/449

BE PR: https://github.com/HSLdevcom/bultti/pull/450

### DIFF
--- a/src/util/userRoles.ts
+++ b/src/util/userRoles.ts
@@ -44,7 +44,6 @@ export function hasAccessRights({
     // If all roles are allowed, check that user role exists. If the user is an operator user,
     // ensure that it has access to the operatorId.
     allowedRoles === 'all' &&
-    user.role !== null &&
     // Either the user is not an operator user...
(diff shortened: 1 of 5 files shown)


src/util/userRoles.ts
```diff
@@ -41,10 +41,10 @@ export function hasAccessRights({
   }
 
   if (
-    // If all roles are allowed, check that the user is not blocked. If the user is an operator user,
+    // If all roles are allowed, check that user role exists. If the user is an operator user,
     // ensure that it has access to the operatorId.
     allowedRoles === 'all' &&
-    user.role !== UserRole.Blocked &&
+    user.role !== null &&
```
danieldunderfelt, line 47: Maybe a bit unnecessary if the role is not nullable.

danieldunderfelt (changes requested):
Small role nullability change, then LGTM!

PR #136, opened 2021-08-25
427 contract refactor

danieldunderfelt:
Closes HSLdevcom/bultti#427

Server PR: https://github.com/HSLdevcom/bultti/pull/466

Remove redundant code and update schema. Add contract assignment to procurement unit config form.


src/text/fi.json
```diff
@@ -38,7 +38,9 @@
   "average": "Keski-arvo",
   "selectAll": "Valitse kaikki",
   "allSelected": "Kaikki valittu",
+  "edited": "Muokattu",
   "procurementUnits": "Kilpailukohteet",
+  "contract": "Sopimusehdot",
```
Divinare, line 43: There was already "contracts": "Sopimusehdot" ^^

Divinare (changes requested):
Looks good, some issues found during testing:

1) Remove test data button doesn't work...
Not sure if it's just me or what's the reason... Well, i'm happy as long as it works for every1 else. Before merging though, I want to be able to remove & create test data in this PR

2) It was forgotten to attach contracts to procurementUnits of testData

3) As I try to remove contract that is being attached to a procurementUnit, I get an error:
update or delete on table "contract" violates foreign key constraint "FK_d9203ebfe3159eb73db72ec667c" on table "procurement_unit"

Shouldn't it work? I think it should. BUT we should prevent removing contracts that are used in inspections, so we need the method guardContractBeingUsedInPreInspection() back that was removed from BE (idea: guard so that even if used in DRAFT inspection, you can't remove?)

4) The form's field jump so that it's quite annoying (static fields are in different spot than editable fields)
Solution: add static text field "Maksimi keski-ikä optiokorotuksilla" and also add static text field of showing currently active "Sopimusehdot"

5) As logged in as Operator or HSL user, i wasn't able to open contracts. Was just seeing a list of juicy buttons but clicking them did nothing at page http://localhost:3001/contract
---> Shouldn't contracts be read only for HSL User? Also Operator might want to see contracts that are currently being applied to procurementUnits of that operator
However I can see the point of hiding contracts from operators; if one accidentally places a wrong contract, suddently the operator can see it.
BUT, should contracts be at least read-only visible for HSL users?

PR #140, opened 2021-09-08
491 fix remove unit option

Divinare:
Closes https://github.com/HSLdevcom/bultti/issues/491

BE PR: https://github.com/HSLdevcom/bultti/pull/504

Also added E2E to test this. Had to also fix a render bug; not showing things if loader is on the screen (otherwise form data is bugged and E2E tests fail)


karppanenr:
OK

PR #141, opened 2021-09-08
456 logging in fixes

Divinare:
Closes https://github.com/HSLdevcom/bultti/issues/456

* First problem solution: remove auth token from clear local storage if login fails so that user does not have to manually reset application memory from browser's console. Hopefully this solves the problem.
* So the second problem was that when logging in and code param started with a number, isNumber thought that the string was a number and tried to parse it. The fix for isNumber check taken from here:
https://stackoverflow.com/a/50376498/4282381


karppanenr:
OK

PR #143, opened 2021-09-23
490 Keski-iän pyöristämisen selvitys

karppanenr:
Closes https://github.com/HSLdevcom/bultti/issues/490

Changed the rounding method to the general round half up method.


Divinare:
Review ok!

